PR #54, opened 2017-07-11
Ft/gf2n: n = 32 & 64

lamphamsy:
# Support GF2N(n=32, 64)

# Operations

| Method | n <= 16 | n > 16 |
| --------| -------- | ------ |
| Multiplication | Log table | Split table |
| Division | Log table | Multiplication + inversion |
| Inversion | Use division | Extended Euclide algorithm |

TODO: division using extended Euclide algorithm

### DIFF
--- a/src/gf2n.h
+++ b/src/gf2n.h
@@ -3,10 +3,6 @@
 
 #include "gf.h"
 
-enum MulType { mul_log_tab, split_8_8 };
-enum DivType { div_log_tab, div_by_inv };
-enum InvType { inv_by_div, inv_ext_gcd };
-
 template<typename T>
 class GF2N : public GF<T>
 {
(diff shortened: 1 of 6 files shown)


vrancurel:
It looks good but can you add tests in tools/test_ec.sh ?

src/gf2n.h
```diff
@@ -3,15 +3,39 @@
 
 #include "gf.h"
 
+enum MulType { mul_log_tab, split_8_8 };
+enum DivType { div_log_tab, div_by_inv };
+enum InvType { inv_by_div, inv_ext_gcd };
+
```
vrancurel: Can you put enum values in upper case, e.g. MUL_LOG_TAB and inside the class definition ?

test/gf_utest.cpp
```diff
@@ -185,6 +220,7 @@ void gf_utest()
   gfutest_uint32.gf_utest();
   GFUtest<uint64_t> gfutest_uint64;
   gfutest_uint64.gf_utest();
+  gfutest_uint64.gf_utest_2_32();
   GFUtest<mpz_class> gfutest_mpz;
   gfutest_mpz.gf_utest_nogf2n(); //XXX gf2n broken for now
 }
```
vrancurel: can you also add a test in tools/test_ec.sh ? e.g. _16 and _32 ?

src/gf2n.h
lamphamsy: yes sure

test/gf_utest.cpp
lamphamsy: yes, I forgot that. thanks

PR #84, opened 2017-09-10
use fft2K in fecfntrs

vrancurel:
don't cache omegas in fft2k (easy to compute for relatively small n)
keep omega caching for schonage-strassen
rename some badly named unit test
add large symbol (encode only) test for showing superiority of butterfly approach


vrancurel:
@lamphamsy look at the test results for RS(50,50)

lamphamsy:
Yes, FFT2K makes a huge improvement, even better than traditional RS codes. We're in a good direction :)

PR #89, opened 2017-09-18
move v2vec in its own header

vrancurel:


vrancurel:
I merged it because it is straightforward

PR #90, opened 2017-09-18
Ft/find highly composite prime minus one

vrancurel:


tools/find_pminus1_compo.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+template class GF<mpz_class>;
+template class GFP<mpz_class>;
+
+int main(int argc, char **argv)
+{
+  GFP<mpz_class> gfp(3);
+
+  if (argc != 3) {
+    std::cerr << "usage: is_prime start_n min_n_factors\n";
```
lamphamsy: It should be
`usage: find_pminus1_compo start_n min_n_factors`

PR #94, opened 2017-09-19
FECGFPFFTRS: GFP(p) based RS

lamphamsy:
Necessary condition: `p >= 2^(8*word_size)` to cover all possible values of symbols, as each symbol is of `(8*word_size)` bits.

A length-`k` vector `X` is encoded into a length-`n` vector `Y` whose elements are GFP elements. Hence, elements of `Y` could be greater than `2^(8*word_size)`.

Note: to facilitate our implementation, we CHOOSE `p < 2 * 2^(8*word_size)`.
We can deal with this as below.
Supposing that an element `Y[i] > 2^(8*word_size)`, we will store its value modulo `2^(8*word_size)`, i.e.

```
Y_p[i] = Y[i] % 2^(8*word_size),
```

a flag will be marked for the position `i`.

To decode `X`, we read `Y[i]` from `Y_p[i]` as:

```
Y[i] = Y_p[i] + 2^(8*word_size)
```
Because `p < 2 * 2^(8*word_size)`, the flag will be only a bool value.

### DIFF
--- a/src/fecgfpfftrs.h
+++ b/src/fecgfpfftrs.h
@@ -2,28 +2,53 @@
 #pragma once
 
 /**
- * GF_2^2^k+1 based RS (Fermat Number Transform)
- * As suggested by the paper:
- * FNT-based Reed-Solomon Erasure Codes
- * by Alexandre Soro and Jerome Lacan
+ * GFP(p) based RS
+ *
+ * Necessary condition: p >= 2^(8*word_size) to cover all possible values of
+ *  symbols, as each symbol is of (8*word_size) bits.
+ *
+ * A length-k vector X is encoded into a length-n vector Y whose elements are
+ *  GFP elements. Hence, elements of Y could be greater than 2^(8*word_size).
+ *  We can deal with this as below.
+ *    Note: to facilitate our implementation, we CHOOSE p < 2 * 2^(8*word_size).
+ *  Supposing that an element Y[i] > 2^(8*word_size), we will store its value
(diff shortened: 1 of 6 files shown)


src/fecgfpfftrs.h
```diff
@@ -0,0 +1,208 @@
...
+  {
+    VVec<T> vwords(words, n);
+    fft->fft(output, &vwords);
+    // check for out of range value in output
+    for (int i = 0; i < n; i++) {
+      if (output->get(i) >= this->limit_value) {
+        char buf[256];
+        snprintf(buf, sizeof (buf), "%lld:%d", offset, i);
+        assert(nullptr != props[i]);
+        props[i]->insert(std::make_pair(buf, "@"));
```
vrancurel: You have to store the value of the number instead of @

src/fecgfpfftrs.h
```diff
@@ -0,0 +1,208 @@
...
+    for (int i = 0; i < k; i++) {
+      vx.set(i, this->gf->exp(r, fragments_ids->get(i)));
+    }
+
+    for (int i = 0; i < k; i++) {
+      int j = fragments_ids->get(i);
+      char buf[256];
+      snprintf(buf, sizeof (buf), "%lld:%d", offset, j);
+      if (nullptr != props[j]) {
+        if (props[j]->is_key(buf))
```
vrancurel: you have to get the value in props

PR #97, opened 2017-09-20
FFT2K: improvement

lamphamsy:
Temporary vectors are defined as the class's members. It improves speed
since their alloc/dealloc occur once instead of in every FFT operation.


lamphamsy:
@vrancurel : it improves ~1.5x speed :)

PR #102, opened 2017-09-29
Ft/cimpl

vrancurel:


src/vec.cpp
```diff
@@ -0,0 +1,217 @@
...
+}
+
+void _vec_hadamard_mul_65537(int n,
+                             uint64_t *x,
+                             uint64_t *y)
+{
+  int i;
+
+  for (i = 0; i < n/2; i++) {
+    x[i] = (((uint64_t) x[i]) * y[i]) % 65537;
```
lamphamsy: x[i] is actually uint64_t, so a convert is not necessary here :)

src/vec.cpp
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  for (int j = 0; j < n/2; j++) {
+    x[i + j] = (((uint64_t) x[i + j]) * y[j]) % 65537;
```
lamphamsy: same here

PR #109, opened 2017-10-06
prepare for GF_p^n

vrancurel:
- create the CG (cyclic group) generic class (superclass of GF)
- vec, mat, and poly are based upon cyclic groups and not fields
- protect p, n in various finite group classes and create accessors (because of symbol conflicts)

### DIFF
--- a/src/fft.h
+++ b/src/fft.h
@@ -9,15 +9,16 @@
 template<typename T>
 class FFT
 {
- public:
+ protected:
   GF<T> *gf;
   Arith<T> *arith;
   int n;
   T inv_n_mod_p;
- protected:
   FFT(GF<T> *gf, int n);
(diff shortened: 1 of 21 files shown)


src/fft.h
```diff
@@ -28,7 +28,8 @@ FFT<T>::FFT(GF<T> *gf, int n)
 {
   this->gf = gf;
   this->n = n;
-  this->inv_n_mod_p = (gf->inv(n) % gf->p);
+  //this->inv_n_mod_p = (gf->inv(n) % gf->card());
```
vrancurel: I think this is gf->card() here, @lamphamsy can you fix ?

lamphamsy: You did it :)

PR #112, opened 2017-10-08
FFTADD: enhancement

lamphamsy:
- Taylor expansion works on vectors
- Improvement of Taylor expansion: computing `g0`, `g1`; multitply f(x) to (x^t - x)

- Vec:  constructor with mem and mem length & copy
    - mem (=NULL by default): an array pointer. If it's given, instance of vector's `mem` is actually this pointer.
    - mem_len (=0 by default): length of `mem`
    - Do copy on `mem` array
        - copy(vec): do copy vec's `mem` to this vector
        - copy(vec, n): do copy vec's `mem` to this vector; zero out next (n-vec's `mem_len`) elements of the vector's `mem` if n > vec's `mem_len`


lamphamsy:
@vrancurel : I add some modifications on vec.h compared to your last review :)

lamphamsy:
@vrancurel : FEC using FFTADD is 2x better than that uses FFTCT, but it's still much worse than FNT :)

PR #113, opened 2017-10-10
Fix/gf2n alloc/dealloc mask

lamphamsy:
Delete `mask` at destructor

- [x] #112


lamphamsy:
These changes will be integrated in #116

PR #117, opened 2017-10-12
prepare for GFPN

vrancurel:
- rename mg to rn (it is not a group it is a ring since it has addition and mul)
- provide converters for polynomial to number representations
- rationalize a bit gf tests
- polynomials can be created over a ring or a field now

### DIFF
--- a/src/gf.h
+++ b/src/gf.h
@@ -11,9 +11,9 @@ class GFP;
  * @param n exponent
  */
 template<typename T>
-class GF : public MG<T>
+class GF : public RN<T>
 {
-private:
+protected:
   T p;
   T n;
   GFP<T> *sub_field;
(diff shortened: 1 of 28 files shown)


src/gf.h
```diff
@@ -22,10 +22,12 @@ class GF : public MG<T>
   GF(T p, T n);
   virtual ~GF();
   GF<T> *get_sub_field();
+  T get_p();
+  T get_n();
 };
 
 template <typename T>
-GF<T>::GF(T p, T n) : MG<T>(this->arith->exp(p, n))
+GF<T>::GF(T p, T n) : RN<T>(this->arith->exp(p, n))
```
lamphamsy: It causes warnings since `arith` has not yet created.

src/poly.h
```diff
@@ -8,12 +8,14 @@ class Vec;
 template<typename T>
 class Poly
 {
- public:
+private:
```
lamphamsy: 👍

PR #120, opened 2017-10-19
Add GFPN class

vrancurel:
- Poly<T> MUST apply to sub_fields of extension fields (not the RING of integers modulo N neither an extension field)
- Add gfpn class with
- Add polynomial gcd and extended euclidean algorithm e.g. to compute the inverse of a polynomial in a field extension
- fix various polynomial functions (e.g. set() when the coef is zero, conversion functions from/to number reprentation)
- add simple extension field tests

### DIFF
--- a/src/gfpn.h
+++ b/src/gfpn.h
@@ -1,0 +1,223 @@
+/* -*- mode: c++ -*- */
+#pragma once
+#include "ntl.h"
+
+template<typename T>
+class Poly;
+
+template<typename T>
+class GFPN : public GF<T>
+{
+private:
+  GF<T> *sub_field;
+  int max_deg;
+  Poly<T> *prim_poly;
+
+public:
+  explicit GFPN(T p, T n);
+  ~GFPN();
+  T card(void);
+  T card_minus_one(void);
+  T neg(T a);
+  T add(T a, T b);
+  T sub(T a, T b);
+  T mul(T a, T b);
+  T div(T a, T b);
+  T inv(T a);
+  T exp(T a, T b);
+  T log(T a, T b);
+  T weak_rand(void);
+};
+
+template <typename T>
+GFPN<T>::GFPN(T p, T n) : GF<T>(p, n)
+{
+  // XXX shall check that p is prime
+  sub_field = new GFP<T>(p);
+  max_deg = 31; //XXX
+
+  //XXX temp
+  if (p == 2) {
+    if (n == 8) {
+      prim_poly = new Poly<T>(sub_field);
+      prim_poly->set(8, 1);
+      prim_poly->set(4, 1);
+      prim_poly->set(3, 1);
+      prim_poly->set(1, 1);
+      prim_poly->set(0, 1);
+    }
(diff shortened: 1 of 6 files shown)


test/gf_utest.cpp
```diff
@@ -139,6 +182,40 @@ class GFUtest
...
+    test_log(&gf9);
+  }
+
+  void test_prime_root_gf9()
+  {
+    std::cout << "test_prime_root_gf9\n";
+
+    GFPN<T> gf9(3, 2);
+    test_find_prime_root(&gf9);
+    //XXX test_get_order(&gf9);
```
vrancurel: @lamphamsy , the get_order() function is probably not well adapted for GF_p^n fields, can you have a look / see how to generalize it ?

vrancurel:
I think now it is possible to write a fecgfpnfftrs for any extension field (e.g. GF_27 !)

test/gf_utest.cpp
lamphamsy: okay, I will do it

lamphamsy: it's done

src/gfpn.h
```diff
@@ -0,0 +1,225 @@
...
+  return _b2.to_num();
+}
+
+template <typename T>
+T GFPN<T>::exp(T a, T b)
+{
+  assert(this->check(a));
+  assert(this->check(b));
+
+  return GF<T>::exp_naive(a, b);
```
lamphamsy: why don't we use exp_quick here?

vrancurel: feel free to change !

src/gfpn.h
```diff
@@ -0,0 +1,225 @@
...
+  assert(this->check(a));
+  assert(this->check(b));
+
+  Poly<T> _a(sub_field);
+  _a.from_num(a, max_deg);
+  Poly<T> _b(sub_field);
+  _b.from_num(b, max_deg);
+
+  _a.add(&_b);
+  _a.mod(prim_poly);
```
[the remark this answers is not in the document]
vrancurel: please remove

src/rn.h
```diff
@@ -522,11 +522,12 @@ template <typename T>
...
-  std::vector<T> primes;
-  std::vector<T> exponent;
   T h = this->card_minus_one();
-  _factor_prime<T>(h, &primes, &exponent);
-  T order = do_step_get_order(x, h, &primes, &exponent);
+  if (!this->compute_factors_of_order_done)
+    _factor_prime<T>(h, this->primes, this->exponent);
+  std::vector<T> _primes(*primes);
+  std::vector<T> _exponent(*exponent);
+  T order = do_step_get_order(x, h, &_primes, &_exponent);
```
vrancurel: @lamphamsy do we need exponent and order as T ? can't just be integers ? (maybe stupid question and we need it)

lamphamsy: Since order could be (q-1), it should be as T. For exponent, I think it could be integers :)

PR #122, opened 2017-10-20
Arith: use exponent as integer

### LINKED ISSUE #121: Don't use T for exponents

lamphamsy:
In prime factoring a number, prime is as T but exponent as integer.
Fix #121


lamphamsy:
Yes, it is. The code is updated.

PR #134, opened 2017-12-29
Enhancement: ngff4 & fec

lamphamsy:
# Enhancement

- NGFF4: refactor
- FEC: use inline for read/write words

# Bench: modify encoding speed formula

Encoding speed is now defined as below

enc_speed = chunk_size * number_coded_chunks / encoding_time

For systematic codes, number_coded_chunks = number of parities
Otherwise, number_coded_chunks = code length
  
# Clean codes

To remove errors caused by using new flags in compiling

### DIFF
--- a/benchmark/Makefile
+++ b/benchmark/Makefile
@@ -1,7 +1,7 @@
 
 #CXX = clang++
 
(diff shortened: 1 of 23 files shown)


lamphamsy:
@vrancurel : it's ready for reviews :)

Issue #155, opened 2018-02-09
Use namespace


slaperche-scality:
Use namespace in the codebase:
We could have the following namespaces:
- nttec
- nttec::base
- nttec::fft
- nttec::fec
- nttec::poly
- nttec::vec

Issue #156, opened 2018-02-09
Better naming scheme for classes


slaperche-scality:
Currently, some classes names are hard to read, this could be improved by making the different part of the name more visible (by using CamelCase, since it's the standard scheme for C++ classes).

For example: "FECGF2NFFTADDRS" vs "FecGf2nFftAddRs", the latter is more easy to read (you can visually see that it's about FEC, GF2N, FFT add and RS).

PR #168, opened 2018-02-12
use namespaces

slaperche-scality:
We introduce the following namespaces:
- nttec: the root namespace of the NTTEC library
- nttec::arith: for the base/core arithmetical functions of NTTEC.
- nttec::fec: for the code related to Forward Error Correction.
- nttec::gf: for the code related to Galois Fields.
- nttec::fft: for the code related to the Fast Fourier Transform.
- nttec::vec: for the low-level wrappers around memory buffer.

Note that we add to move the overloading of << for the 128-bit integers
into the namespace "std", otherwise we have some problem to find it when
needed (ADL cannot find it in some cases).

Also move pack/unpack helper of Vec from `fec.h` to `vec.h`: it's more
logical that way.

Closes: #155

### DIFF
--- a/benchmark/benchmark.cpp
+++ b/benchmark/benchmark.cpp
@@ -264,7 +264,7 @@ int Benchmark<T>::check_params()
         }
     }
 
-    size_t wordsize_limit = nttec::arith::_log2<T>(n) + 1;
+    size_t wordsize_limit = nttec::arith::log2<T>(n) + 1;
     if (wordsize_limit > 8 * word_size) {
         return ERR_COMPT_CODE_LEN_T;
     }
--- a/benchmark/benchmark.cpp
+++ b/benchmark/benchmark.cpp
@@ -124,31 +124,33 @@ int Benchmark<T>::init()
 {
     switch (fec_type) {
     case EC_TYPE_GF2NRSV:
-        fec = new FECGF2NRS<T>(word_size, k, m, FECGF2NRS<T>::VANDERMONDE);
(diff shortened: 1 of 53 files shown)


slaperche-scality:
I put things in namespace as I understood it, but since I don't have a complete grasp of each class yet it wouldn't hurt if someone else could do a quick check.

Like I put NGFF4 in the gf namespace, but NGFF4 is not a Galois Field so maybe:
- it's fine, but rename the namespace
- if's fine, but update the Doxygen of the namespace
- it's not fine: create a new namespace/move it into another one

vrancurel:
I think it is fine if we have n*GF(F_4) in gf namespace

benchmark/benchmark.cpp
```diff
@@ -262,7 +264,7 @@ int Benchmark<T>::check_params()
         }
     }
 
-    size_t wordsize_limit = _log2<T>(n) + 1;
+    size_t wordsize_limit = nttec::arith::_log2<T>(n) + 1;
```
vrancurel: Now we have proper namespaces we can now remove underscores from arith functions (as they were used to distinguish them mainly from classes methods).

slaperche-scality: Yes, that's another patch I already have
I didn't put it in this PR because I think it was more related to #156.
That being said, arith is not a class so it makes more sense to add it in this PR.

lamphamsy:
I do agree to put ngff4 in GF namespace since we are using ngff4 as other GFs :)

PR #172, opened 2018-02-14
export the version string

slaperche-scality:
I'll use this function to test the inclusion of NTTEC in QuadIron's build system.


slaperche-scality:
Only for review now, I have a PR in progress in QuadIron and there is still some stuff to be done (either here or there).
Until everything is figured out, we shouldn't merge this one in case there is still some adjustement to be done on this side.

PR #173, opened 2018-02-14
rework KeyValue

slaperche-scality:
The was room for improvement for KeyValue:
1. the name is misleading: a KeyValue object represents a set of
   key/value pairs, not a single pne
2. we use std::map, but we don't care about the ordering so
   std::unordered_map is more efficient/logical
3. we use inheritance instead of composition, this is overkill since we
   only need a subset of the associative container API for our object.
4. we use string (built from integers) as key: we have to do string
   formatting (costly) for each lookup
5. accessing a value requires two lookup (`count` to check if the key
   exists + the actual lookup) : this can be done with a single lookup!


benchmark/benchmark.cpp
```diff
@@ -423,7 +420,7 @@ void Benchmark<T>::get_avail_chunks(
         for (i = 0; i < k; i++) {
             int j = c_chunks_id->at(i);
             avail_c_chunks->at(j) = a_streams->at(j);
-            avail_c_props->at(j) = c_propos->at(j);
+            avail_c_props.at(j) = c_propos.at(j);
```
vrancurel: is c_propos a typo ? should be c_props ?

slaperche-scality: Yes, could be.
Typo seems also present in `fec_utest.cpp` (`propos_flags`), I'll fix it in a dedicated commit (but in this PR)

lamphamsy: Yes, it is :)

PR #179, opened 2018-02-26
Renaming & co

slaperche-scality:
The goal of this PR is to improve readability.
This should be the last "cataclysmic" PR that touch everything but the kitchen sink (other changes such as adding lint or refactoring code will be more incremental).

This PR:
- formats the formulas using LaTeX (I only did top level ones, there may be other to do)
- adds some "schemas" when needed (to show the differences between all the Vec* implementations)
- cites research papers using BibTex  (I only did top level ones, there may be other to do)
- rename classes/files to be more readable for the novice (s/fecgf2nfftaddrs/fec_rs_gf2n_fft_add/ => using _ make the acronyms more visible, s/VmVec/vec::Slice/ is more understandable, …)


src/vec_buf_doubled.h
```diff
@@ -0,0 +1,69 @@
...
+namespace vec {
+
+/** A vector of `n` buffers virtually extented to `2n`.
+ *
+ * Physically, BuffersDoubled contains `n` buffers but behaves as if it
+ * contained `2n` buffers.
+ *
+ * Example:
+ *
+ * We have a set of `N` independent buffers (+ a buffer of zeros):
```
lamphamsy: `(+ a buffer of zeros)` should not be there, shouldn't it?

src/vec_buffers.h
```diff
@@ -11,34 +11,38 @@
...
- * Wrapper of a vector of `n` buffers each of `size` elements of type T
- *  std::vector<T*> vec(n)
- *    vec[i] = T[size];
+/** A vector of `n` buffers (array of T).
+ *
+ * A Buffers contains pointers to `n` buffers such as
+ *
+ * Example:
+ *
+ * We have a set of `N` independent buffers (+ a buffer of zeros):
```
lamphamsy: `(+ a buffer of zeros)` should not be there, shouldn't it?

src/vec_buf_doubled.h
slaperche-scality: Good catch!

Too much copy/paste

src/vec_buffers.h
slaperche-scality: Yup, I'll fix this.

PR #190, opened 2018-03-05
SIMD: add vectorized functions

lamphamsy:
Currently, it supports only for CPUID SSE4.1
Supported components:
- gf::nf4 and fec::rs-nf4 for (w=8, t=16)
- vec::Buffers and fec::fnt for (w=1,2 and t=4)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,7 +144,7 @@ endforeach(flag)
 # Release flags
 # Debug flags
 set(RELEASE_CXX_FLAGS
-  -O3 -flto -DNDEBUG
+  -O3 -DNDEBUG
   -msse4.1
 )
 foreach(flag ${RELEASE_CXX_FLAGS})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,7 +142,12 @@ foreach(flag ${DEBUG_CXX_FLAGS})
 endforeach(flag)
 
 # Release flags
-foreach(flag -O3 -flto -DNDEBUG)
+# Debug flags
+set(RELEASE_CXX_FLAGS
+  -O3 -flto -DNDEBUG
+  -msse4.1
+)
+foreach(flag ${RELEASE_CXX_FLAGS})
   check_cxx_compiler_flag(${flag} has_flag_${flag})
   if(has_flag_${flag})
     set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} ${flag}")
--- a/src/simd.h
+++ b/src/simd.h
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2017-2018 the NTTEC authors
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __NTTEC_SIMD_H__
+#define __NTTEC_SIMD_H__
+
+// FIXME: detect SIMD flags since it requires SSE4.1 instructions
+#include "simd_128.h"
+
+#endif
--- a/src/simd_128.h
+++ b/src/simd_128.h
@@ -1,0 +1,388 @@
+/*
+ * Copyright 2017-2018 the NTTEC authors
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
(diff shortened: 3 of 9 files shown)


slaperche-scality:
I don't know why the tests are failing, but for the lint the errors comes from the static analyser of Clang:

```
nttec/src/simd_128.h:65:5: warning: Undefined or garbage value returned to caller [clang-analyzer-core.uninitialized.UndefReturn]
nttec/src/simd_128.h:224:20: warning: 1st function call argument is an uninitialized value [clang-analyzer-core.CallAndMessage]
nttec/src/simd_128.h:65:5: warning: Undefined or garbage value returned to caller [clang-analyzer-core.uninitialized.UndefReturn]
nttec/src/simd_128.h:99:5: warning: Undefined or garbage value returned to caller [clang-analyzer-core.uninitialized.UndefReturn]
nttec/src/simd_128.h:129:5: warning: Undefined or garbage value returned to caller [clang-analyzer-core.uninitialized.UndefReturn]
nttec/src/simd_128.h:265:5: warning: Undefined or garbage value returned to caller [clang-analyzer-core.uninitialized.UndefReturn]
nttec/src/simd_128.h:273:5: warning: Undefined or garbage value returned to caller [clang-analyzer-core.uninitialized.UndefReturn]
nttec/src/simd_128.h:321:5: warning: Undefined or garbage value returned to caller [clang-analyzer-core.uninitialized.UndefReturn]
nttec/src/simd_128.h:353:5: warning: Undefined or garbage value returned to caller [clang-analyzer-core.uninitialized.UndefReturn]
```
This looks like real bug (except if they are false positives).
Running a recent clang (5.0) locally, you can get more information (such as the codepath followed by the analyser step by step).

lamphamsy:
@slaperche-scality : these warning come from the way of getting value from register, e.g.
```
    __uint128_t c;
    _mm_storeu_si128((__m128i*)&c, b);
```
I will resolve them.

BTW, tests failed as some functions (expand16/32) were implemented for (word_size=4, T=128bit)

CMakeLists.txt
```diff
@@ -115,6 +115,7 @@ include(CheckCXXCompilerFlag)
 set(COMMON_CXX_FLAGS
   -pipe
   -Wall
+  -march=native
```
slaperche-scality: This should be for release flags only, not the common ones.

Also, we may want to use `-msse4.1` instead of `-march=native` here:  if you use `-march=native` then the binary compiled for one machine may not run on another machine.

CMakeLists.txt
```diff
@@ -142,7 +143,7 @@ foreach(flag ${DEBUG_CXX_FLAGS})
 endforeach(flag)
 
 # Release flags
-foreach(flag -O3 -flto -DNDEBUG)
```
slaperche-scality: Was there some issue with LTO?

src/simd.h
```diff
@@ -0,0 +1,80 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __NTTEC_SIMD_H__
+#define __NTTEC_SIMD_H__
+
+#ifdef _MSC_VER
```
slaperche-scality: Do we really support Microsoft compiler?

src/simd_128.h
```diff
@@ -0,0 +1,381 @@
...
+ *
+ *  Currently, it supports operations on 32-bit numbers
+ */
+namespace simd {
+
+/* ==================== Essential Operations =================== */
+
+const uint32_t f4 = 65537;
+const uint32_t f3 = 257;
+const __m128i f4_tuple = _mm_set1_epi32(65537);
```
slaperche-scality: You could use your constant here

```
const uint32_t F4 = 65537;
const uint32_t F3 = 257;
const __m128i F4_x4 = _mm_set1_epi32(F4);
const __m128i F4minus1_x4 = _mm_set1_epi32(F4 - 1);
const __m128i F3_x4 = _mm_set1_epi32(F3);
const __m128i F3minus1_x4 = _mm_set1_epi32(F3 - 1);
```

CMakeLists.txt
lamphamsy: Yes, it should be for release flags.
I use the flag `march=native` instead of specific flags to let compiler uses the most efficient one. `simd.h` will choose the appropriated files, e.g. `simd_128.h` for now, or `simd_256.h` (later) for AVX2 etc.

src/simd.h
lamphamsy: I will clean it :)

CMakeLists.txt
slaperche-scality: I think we will need a proper framework to select the most-efficient one (and to have a fallback for CPU with no SIMD), it won't be solved by `march=native` alone.
It's better to select the most optimized version dynamically (using CPU dispatch once, at the lib initialization) rather than using compile-time settings (that's how most of Intel libs works).
We don't have such "framework" for now, but we should introduce one at one point.

IMHO, if we want to compile with `march=native` it should be done by enabling an option.
Generating non-portable binary by default hurt the reproducibility of our benchmark/tests/etc.
It can be OK for local dev/test, but no good for CI & cie.

CMakeLists.txt
lamphamsy: oh, there was error for previous versions. But it works fine now. I will remove this change.

CMakeLists.txt
lamphamsy: I like that framework :)
So maybe we should have an option to generating binaries with SIMD instruction?

slaperche-scality: Yes, we could add an option to CMake (USE_SIMD, or USE_HW_ACCELERATION) that would be True by default (but could be disabled on-demand, if people compile on old/unsupported CPU).
When this option is set, we would have a preprocessor symbol defined, and then we can use #ifdef in our code to include or not SIMD code according to this.

Latter, we can add a "framework" for runtime selection, but anyway SIMD should be optional so we will need a compile-time flag for it (can be done in another PR though, this PR is mainly about introducing your SIMD code).

slaperche-scality:
About the static analyser warnings, it seems to have issue with the pattern:

```
__uint128_t c;
_mm_storeu_si128((__m128i*)&c, b);
return c;
```

It seems like it can't see that `_mm_storeu_si128` will initialize the value, so it's a false positive.
What I propose is to define a helper that will encapsulate this pattern and disable the lint for this helper and then use this helper in the code:

```
diff --git a/src/simd_128.h b/src/simd_128.h
index 97f91d0..6988a51 100644
--- a/src/simd_128.h
+++ b/src/simd_128.h
@@ -51,6 +51,13 @@ const __m128i f4minus1_tuple = _mm_set1_epi32(65536);
 const __m128i f3_tuple = _mm_set1_epi32(257);
 const __m128i f3minus1_tuple = _mm_set1_epi32(256);
 
+static inline __uint128_t m128i_to_uint128(__m128i v)
+{
+    __uint128_t i;
+    _mm_storeu_si128((__m128i*)&i, v);
+    return i;  // NOLINT(clang-analyzer-core.uninitialized.UndefReturn)
+}
+
 /** Perform a%card where a is a addition of two numbers whose elements are
  *  symbols of GF(card) */
 inline __uint128_t mod_after_add(__m128i a, uint32_t card)
@@ -60,9 +67,8 @@ inline __uint128_t mod_after_add(__m128i a, uint32_t card)
 
     __m128i cmp = _mm_cmpgt_epi32(a, _card_minus_1);
     __m128i b = _mm_sub_epi32(a, _mm_and_si128(_card, cmp));
-    __uint128_t c;
-    _mm_storeu_si128((__m128i*)&c, b);
-    return c;
+
+    return m128i_to_uint128(b);
 }
 
 /** Perform addition of two numbers a, b whose elements are of GF(card) */
@@ -94,9 +100,7 @@ inline __uint128_t sub(__uint128_t a, __uint128_t b, uint32_t card = f4)
 
     __m128i c = _mm_sub_epi32(_a1, _b);
 
-    __uint128_t _c;
-    _mm_storeu_si128((__m128i*)&_c, c);
-    return _c;
+    return m128i_to_uint128(c);
 }
 
 /** Perform a%card where a is a multiplication of two numbers whose elements are
@@ -124,9 +128,7 @@ inline __uint128_t mod_after_multiply_f4(__m128i a)
 
     __m128i b = _mm_sub_epi32(_lo, hi);
 
-    __uint128_t c;
-    _mm_storeu_si128((__m128i*)&c, b);
-    return c;
+    return m128i_to_uint128(b);
 }
 
 inline __uint128_t mod_after_multiply_f3(__m128i a)
@@ -260,17 +262,15 @@ inline __uint128_t expand16(uint16_t* arr)
 {
     __m128i b = _mm_set_epi64(
         _mm_setzero_si64(), _mm_set_pi16(arr[3], arr[2], arr[1], arr[0]));
-    __uint128_t c;
-    _mm_storeu_si128((__m128i*)&c, b);
-    return c;
+
+    return m128i_to_uint128(b);
 }
 
 inline __uint128_t expand32(uint32_t* arr)
 {
     __m128i b = _mm_set_epi32(arr[3], arr[2], arr[1], arr[0]);
-    __uint128_t c;
-    _mm_storeu_si128((__m128i*)&c, b);
-    return c;
+
+    return m128i_to_uint128(b);
 }
 
 /** Add buffer `y` to two halves of `x`. `x` is of length `n` */
@@ -316,9 +316,8 @@ inline __uint128_t pack(__uint128_t a)
         _mm_extract_epi16(_a, 2),
         _mm_extract_epi16(_a, 1),
         _mm_extract_epi16(_a, 0));
-    __uint128_t c;
-    _mm_storeu_si128((__m128i*)&c, b);
-    return c;
+
+    return m128i_to_uint128(b);
 }
 
 inline __uint128_t pack(__uint128_t a, uint32_t flag)
@@ -348,9 +347,7 @@ inline __uint128_t pack(__uint128_t a, uint32_t flag)
 
     __m128i b = _mm_set_epi32(b3, b2, b1, b0);
 
-    __uint128_t c;
-    _mm_storeu_si128((__m128i*)&c, b);
-    return c;
+    return m128i_to_uint128(b);
 }
 
 } // namespace simd
```

lamphamsy:
@slaperche-scality : thanks so much for the diff 👍

lamphamsy:
I just squashed commits

slaperche-scality:
The bug was due to buggy codegen from the Apple-flavor of GCC (see https://github.com/scality/nttec/pull/192 for some details).

CMakeLists.txt
lamphamsy: In fact we remove the flag because LTO does not work correctly with Clang

PR #194, opened 2018-03-12
SIMD: vectorize subtraction and more

lamphamsy:
- Vectorize subtraction operations => double speed :)
- Add option to enable/disable SIMD vectorizations
- Define alignment types
- Benchmark script: add specific scenarios for fnt and nf4
- circleci: run benchmark with and without SIMD

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,12 @@ set(VERSION ${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}${VERSION_REV})
 set(PACKAGE ${CMAKE_PROJECT_NAME})
 string(TOUPPER "${CMAKE_PROJECT_NAME}" UPCASE_PROJECT_NAME)
 
+# Top level cmake file, set options
+option(NTTEC_USE_SIMD
+  "Build with SIMD vectorizations"
+  OFF
+)
+
 ################
 # Useful defines
 ################
@@ -117,6 +123,12 @@ set(COMMON_CXX_FLAGS
   -Wall
 )
 
+# Option for enabling/disabling SIMD flags is for both of debug and release
+if (NTTEC_USE_SIMD)
+  SET(COMMON_CXX_FLAGS "${COMMON_CXX_FLAGS} -msse4.1")
+  add_definitions(-DNTTEC_USE_SIMD)
+endif()
+
 # Manually add -Werror, for some reasons I can't make it works in the foreach…
 if (CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
   add_compile_options(-Werror)
@@ -142,10 +154,8 @@ foreach(flag ${DEBUG_CXX_FLAGS})
 endforeach(flag)
 
 # Release flags
-# Debug flags
 set(RELEASE_CXX_FLAGS
   -O3 -DNDEBUG
-  -msse4.1
 )
 foreach(flag ${RELEASE_CXX_FLAGS})
   check_cxx_compiler_flag(${flag} has_flag_${flag})
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -42,23 +42,38 @@ then
     exit 1;
 fi
 
-for ec_type in rs-gf2n-v rs-gf2n-c rs-gf2n-fft rs-gf2n-fft-add rs-gfp-fft rs-fnt rs-nf4; do
-  for k in 5; do
-    for m in 2; do
-      for word_size in 1 2 4 8; do
-        ${bin} -e ${ec_type} -w ${word_size} -k ${k} -m ${m} -c ${chunk_size} -s ${sce_type} -g ${threads_nb} -f ${show_type}
-        show_type=0
-      done
+chunk_size=51200
+sce_type=enc_only
+# for rs-nf4
+ec_type=rs-nf4
+word_size=8
+for k in 16; do
+  for n in 256 1024; do
+    m=$((n-k))
+    ${bin} -e ${ec_type} -w ${word_size} -k ${k} -m ${m} -c ${chunk_size} -s ${sce_type} -g ${threads_nb} -f ${show_type}
+    show_type=0
+  done
+done
+
+# for rs-fnt with different packet sizes
+ec_type=rs-fnt
+word_size=2
+for k in 16; do
+  for n in 256 1024; do
+    m=$((n-k))
+    for pkt_size in 0 256 512; do
+      ${bin} -e ${ec_type} -w ${word_size} -k ${k} -m ${m} -c ${chunk_size} -s ${sce_type} -g ${threads_nb} -f ${show_type} -p ${pkt_size}
     done
   done
 done
 
-chunk_size=51200
-sce_type=enc_only
-for word_size in 2 8; do
-  for ec_type in rs-nf4 rs-fnt rs-gfp-fft; do
-    for k in 16; do
-      for m in 64; do
+# for all cases
+sce_type=enc_dec
+chunk_size=512
+for ec_type in rs-gf2n-v rs-gf2n-c rs-gf2n-fft rs-gf2n-fft-add rs-gfp-fft rs-fnt rs-nf4; do
+  for k in 5; do
+    for m in 2; do
+      for word_size in 1 2 4 8; do
         ${bin} -e ${ec_type} -w ${word_size} -k ${k} -m ${m} -c ${chunk_size} -s ${sce_type} -g ${threads_nb} -f ${show_type}
       done
     done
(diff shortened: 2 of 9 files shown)


scripts/benchmark.sh
```diff
@@ -42,23 +42,38 @@ then
...
-      for word_size in 1 2 4 8; do
-        ${bin} -e ${ec_type} -w ${word_size} -k ${k} -m ${m} -c ${chunk_size} -s ${sce_type} -g ${threads_nb} -f ${show_type}
-        show_type=0
-      done
+chunk_size=51200
+sce_type=enc_only
+# for rs-nf4
+ec_type=rs-nf4
+word_size=8
+for k in 16; do
```
slaperche-scality: Do we plan to add more values for `k` later?
If not, we can remove the loop to use k=16 and save one level of indentation.

Same for l.61 and 74-75

src/gf_nf4.h
```diff
@@ -507,7 +507,19 @@ void NF4<__uint128_t>::hadamard_mul(int n, __uint128_t* x, __uint128_t* y);
 template <typename T>
 void NF4<T>::hadamard_mul(int n, T* x, T* y)
 {
-    return;
+    int i;
+    int half = n / 2;
```
slaperche-scality: can be const

src/gf_nf4.h
```diff
@@ -518,7 +530,19 @@ void NF4<__uint128_t>::add(int n, __uint128_t* x, __uint128_t* y);
 template <typename T>
 void NF4<T>::add(int n, T* x, T* y)
 {
-    return;
+    int i;
+    int half = n / 2;
```
slaperche-scality: can be const

src/gf_nf4.h
```diff
@@ -507,7 +507,19 @@ void NF4<__uint128_t>::hadamard_mul(int n, __uint128_t* x, __uint128_t* y);
 template <typename T>
 void NF4<T>::hadamard_mul(int n, T* x, T* y)
 {
-    return;
+    int i;
```
slaperche-scality: You could reduce the scope of `i` by declaring it in the loop directly (same remark for `add`)

src/gf_nf4.h
```diff
@@ -122,8 +122,10 @@ bool NF4<T>::check_n(unsigned n)
     return (n <= sizeof(T) / 4);
 }
 
+#ifdef NTTEC_USE_SIMD
```
slaperche-scality: Maybe we should group all the SIMD specialization declaration in order to have a single `ifdef` block in this file. That would be more readable/easier to maintain.

src/gf_ring.h
```diff
@@ -349,12 +349,15 @@ void RingModN<T>::mul_coef_to_buf(T a, T* src, T* dest, size_t len)
         dest[i] = T((coef * src[i]) % this->_card);
     }
 }
+
```
slaperche-scality: Maybe we should group all the SIMD specialization declaration in order to have a single ifdef block in this file. That would be more readable/easier to maintain.

src/simd.h
```diff
@@ -31,6 +31,8 @@
 #ifndef __NTTEC_SIMD_H__
 #define __NTTEC_SIMD_H__
 
+#ifdef NTTEC_USE_SIMD
```
slaperche-scality: Or maybe we could use

#ifndef NTTEC_USE_SIMD
#error "cannot include simd.h if SIMD is not enabled".
#endif // #ifndef NTTEC_USE_SIMD

src/simd_128.h
```diff
@@ -252,6 +252,35 @@ add_two_bufs(uint32_t* src, uint32_t* dest, size_t len, uint32_t card = F4)
...
 }
 
+inline void sub_two_bufs(
+    uint32_t* bufa,
+    uint32_t* bufb,
+    uint32_t* res,
+    size_t len,
+    uint32_t card = F4)
+{
+    __uint128_t* _bufa = static_cast<__uint128_t*>(static_cast<void*>(bufa));
```
slaperche-scality: Why this double cast?

src/simd_128.h
```diff
@@ -252,6 +252,35 @@ add_two_bufs(uint32_t* src, uint32_t* dest, size_t len, uint32_t card = F4)
...
+    __uint128_t* _bufb = static_cast<__uint128_t*>(static_cast<void*>(bufb));
+    __uint128_t* _res = static_cast<__uint128_t*>(static_cast<void*>(res));
+    size_t _len = len / 4;
```
slaperche-scality: Maybe we could replace 4 (which is "hardcoded") by something like `sizeof(bufb[0])/sizeof(bufb[0]))` or `sizeof(__uint128_t)/sizeof(uint32_t)` and store the result in a named constant that we can use hither and yon.

scripts/benchmark.sh
lamphamsy: I'd like to keep this for loop to facilitate extra runs :)
In CI, I define only a typical value for test :)

src/simd.h
lamphamsy: NTTEC_USE_SIMD can be undefined. It's not an error.

slaperche-scality: Depends how you see things, but IMHO including simd.h with simd disabled could be considered as an error, no?

src/gf_nf4.h
lamphamsy: in fact `i` appears in two loops so I keep it here.

slaperche-scality: But it's not the same `i` (it is set to 0 for each loop), it makes more sense to declare it inline

for (int i = 0; …)

It help the reader (and maybe the compiler) to see the scope of the variable.

src/simd.h
slaperche-scality: But we can keep it like that as well.

There are two approach:
- you always include simd.h, but simd.h is potentially empty (if SIMD is disabled)
- the content of simd.h is always the same, and the other files include it conditionnally

e.g: 

```
#ifdef NTTEC_USE_SIMD
#include "simd.h"
#endif NTTEC_USE_SIMD
```
In gf_nf4.cpp.

The "advantage" of the second approach is that simd.h content never change and anyway we already have a test on NTTEC_USE_SIMD in gf_nf4.cpp so the include could be grouped with the existing one.

lamphamsy: I thought also your second way. It's actually updated in this way. Thanks :)

src/simd_128.h
lamphamsy: Some functions (as this one) would be generalized using template. I will use this calculation in the generalized functions.

src/gf_nf4.h
lamphamsy: it's done

src/simd_128.h
slaperche-scality: Once you cast in void*, you can static_cast in whatever your want, there is no more safety (you tricked the compiler). At this point, a reinterpret_cast is more readable/explicit and as the same level of safety.

lamphamsy: it's done

src/simd_128.h
```diff
@@ -57,7 +57,7 @@ const am128i F3minus1_m128i = _mm_set1_epi32(256);
 static inline aint128 m128i_to_uint128(am128i v)
 {
     aint128 i;
-    _mm_storeu_si128((am128i*)&i, v);
+    _mm_store_si128((am128i*)&i, v);
```
slaperche-scality: Do we have performance improvment thanks to this :slightly_smiling_face: ?
If so, how much?

CMakeLists.txt
```diff
@@ -117,6 +123,14 @@ set(COMMON_CXX_FLAGS
   -Wall
 )
 
+# Option for enabling/disabling SIMD flags is for both of debug and release
+if (NTTEC_USE_SIMD)
+  SET(COMMON_CXX_FLAGS "${COMMON_CXX_FLAGS} -msse4.1")
+  add_definitions(-DNTTEC_USE_SIMD)
+else()
+  set(COMMON_CXX_FLAGS "${COMMON_CXX_FLAGS} -mno-sse4.1")
```
slaperche-scality: I think this is not needed. In fact I'm worried that it can disable auto-vectorisation from the compiler.
USE_SIMD should be used to control our explicit use of SIMD, if set we add the required flag, it not set we compile as-is, no?

src/simd_128.h
lamphamsy: Yes, it improves but not much. But we should work on aligned memory for long term.

CMakeLists.txt
lamphamsy: ah yes, you're right. We have flag to exclude simd codes in the case. Thanks

lamphamsy: I noticed also that -mno-sse leads to a failed compilation due to `llvm/5.0.1/include/c++/v1/unordered_map` and `llvm/5.0.1/include/c++/v1/__hash_table:1161:10`

Issue #68, opened 2017-09-03
move Schönhage-Strassen in a real utils.h function


vrancurel:
https://github.com/vrancurel/ntl/blob/ad47cc23d91abc1bfc20bce8b493db0b6151a4df/test/fft_utest.cpp#L134

vrancurel:
could be useful one day for large polynomials multiplication (decode)

slaperche-scality:
With the removal of GMP in https://github.com/scality/nttec/pull/195, I'm not sure if this issue is still relevant.

slaperche-scality:
Schönhage-Strassen should be reimplement without GMP and as a standalone function.

PR #195, opened 2018-03-13
Remove GMP

slaperche-scality:
It's mainly used by the tools, and it may not play well with the BSD licensing we want for the library itself.


test/fft_utest.cpp
```diff
@@ -103,130 +103,6 @@ class FFTUtest {
         return vec;
     }
 
-    /**
```
slaperche-scality: I think that this makes https://github.com/scality/nttec/issues/68 outdated, if so we could close it.

test/fft_utest.cpp
```diff
@@ -103,130 +103,6 @@ class FFTUtest {
...
-
-        // carry propagation
-        mpz_class z = 0;
-        mpz_class u;
-        for (int i = 0; i <= fft.get_n() - 1; i++) {
-            mpz_class t, b;
-            b = 10;
-            mpz_pow_ui(t.get_mpz_t(), b.get_mpz_t(), i);
-            u = mpz_class(std::to_string(sfXY->get(i)));
-            z += u * t;
```
vrancurel: Can you adapt schonage-strassen in a nttec:utils class/function (we will need it for interpolation probably) and replace carry propagation with GMP by scalar based interpolation ?

slaperche-scality: Ok, I'll add another patch in this PR (and I'll update the description in https://github.com/scality/nttec/issues/68).

vrancurel:
OK so Schonage-Strassen is less urgent ...

lamphamsy:
@slaperche-scality : could you please rebase the branch? :)

slaperche-scality:
@lamphamsy done :)

PR #200, opened 2018-03-23
Polynomial enhancement

### LINKED ISSUE #124: We should operate only on sub_fields
https://github.com/vrancurel/ntl/blob/78f391b2753649de1a488fcca7b36ac99c10c974/src/poly.h#L14

so this notion should disappear, especially gf2n should not create a poly with itselg (GF(2^n)) but with GF(2)

One day we should assert that the field is a prime field

### LINKED ISSUE #167: Factorize Lagrange interpolation code
The code is duplicated in all fec*

lamphamsy:
It does some changes in Polynomial:
- Degree is cached as m_degree. It is updated in set() and copy()
- Terms can be return by get_terms()
- Set() erases only coefficient of degree = m_degree
- Improve neg() and derivative()
- Add two special multiplications: 
   - Multiplication with known max degree of output
   - Multiplication with (X + coef)

~It updates decode of RS-FNT for new polynomial~
It refactor decoding for all FECs

Close #124 
Fix #167

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -183,7 +183,7 @@ coverage_evaluate()
 
 # Be sure to recompile the test driver if needed before running the test suite.
 add_custom_target(check
-  COMMAND ${CMAKE_CTEST_COMMAND}
+  COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure
   DEPENDS ${UNIT_TESTS}
   DEPENDS ${EC_DRIVER}
 )
--- a/src/fec_base.h
+++ b/src/fec_base.h
@@ -34,10 +34,11 @@
 #include <algorithm>
 #include <cassert>
 #include <cstdint>
+#include <memory>
 #include <vector>
-
 #include <sys/time.h>
 
+#include "fft_base.h"
 #include "gf_base.h"
 #include "misc.h"
 #include "property.h"
@@ -163,7 +164,7 @@ class FecCode {
         const std::vector<Properties>& props,
         off_t offset,
         vec::Vector<T>* fragments_ids,
-        vec::Vector<T>* words) = 0;
+        vec::Vector<T>* words);
 
     bool readw(T* ptr, std::istream* stream);
     bool writew(T val, std::ostream* stream);
@@ -207,7 +208,29 @@ class FecCode {
     }
 
   protected:
+    // primitive nth root of unity
+    T r;
     gf::Field<T>* gf;
+    fft::FourierTransform<T>* fft = nullptr;
+    // This vector MUST be initialized by derived Class using multiplicative FFT
+    std::unique_ptr<vec::Vector<T>> inv_r_powers = nullptr;
+
+    virtual void decode_prepare(
+        const std::vector<Properties>& props,
+        off_t offset,
+        vec::Vector<T>* fragments_ids,
+        vec::Vector<T>* words,
+        vec::Vector<T>* vx,
+        int* vx_zero = 0);
+
+    virtual void decode_Lagrange(
+        vec::Vector<T>* output,
+        const std::vector<Properties>& props,
+        off_t offset,
+        vec::Vector<T>* fragments_ids,
+        vec::Vector<T>* words,
+        vec::Vector<T>* vx,
+        int vx_zero = 0);
 };
 
 /**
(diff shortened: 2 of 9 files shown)


src/polynomial.h
```diff
@@ -176,23 +176,19 @@ void Polynomial<T>::set(int exponent, T coef)
...
 
-    typename Term<T>::const_iterator it = terms.find(exponent);
-
+    auto it = terms.find(exponent);
     if (it == terms.end()) {
         if (coef == 0)
             return;
     } else {
-        if (coef == 0) {
+        if (coef == 0 && exponent == _degree) {
```
slaperche-scality: Why these changes? You could add some explanation in the commit message?

It super useful when you git blame some code to have some context/explanation in the commit message.

src/polynomial.h
```diff
@@ -111,6 +112,12 @@ Polynomial<T>::Polynomial(gf::Field<T>* field)
     this->_degree = 0;
 }
 
+template <typename T>
+Polynomial<T>::~Polynomial()
+{
+    clear();
```
slaperche-scality: Is it useful?

`clear` does two things:

- set `_degree` to 0: useless, the object is being destroyed
- calling `clear` on `terms`: IIRC `terms` is a map, so its destructor will be called here, there is no need to clear it beforehand, no?

src/polynomial.h
```diff
@@ -60,7 +60,7 @@ class Polynomial {
     void copy(Polynomial<T>* src);
     void copy(Polynomial<T>* src, T offset);
     int degree();
-    Term<T> get_terms();
+    Term<T>* get_terms();
```
slaperche-scality: Do you need a write access to `terms`? Otherwise you could return a const reference (and declare the method as **const**)

src/polynomial.h
```diff
@@ -86,6 +86,8 @@ class Polynomial {
     void add(Polynomial<T>* b);
     void sub(Polynomial<T>* b);
     void mul(Polynomial<T>* b);
+    void mul(Polynomial<T>* b, int deg_out);
```
slaperche-scality: Commit should explain why we need these specialization/how they improve the perf.

src/polynomial.h
```diff
@@ -438,6 +440,57 @@ void Polynomial<T>::mul(Polynomial<T>* b)
     _mul(this, &a, b);
 }
 
+/** Multiply to a polynomial b. Degree of result polynomial is limited by a
+ * given value, i.e. elements of degree greater than this value are ignored.
+ *
+ * @param b - polynomial to multiply
+ * @param deg_out - degree of result
+ * @return
```
slaperche-scality: If there is no return, you can remove the `@return`

src/polynomial.h
```diff
@@ -438,6 +440,57 @@ void Polynomial<T>::mul(Polynomial<T>* b)
     _mul(this, &a, b);
 }
 
+/** Multiply to a polynomial b. Degree of result polynomial is limited by a
```
slaperche-scality: To allow Doxygen to extract a short description, this should be formatted as

```
/** Multiply to a polynomial b.
 *
 * Degree of result polynomial is limited by a given value, i.e. elements of
 * degree greater than this value are ignored.
```

src/polynomial.h
```diff
@@ -438,6 +440,57 @@ void Polynomial<T>::mul(Polynomial<T>* b)
...
+ * @param deg_out - degree of result
+ * @return
+ */
+template <typename T>
+void Polynomial<T>::mul(Polynomial<T>* b, int deg_out)
+{
+    for (int deg = deg_out; deg >= 0; deg--) {
+        T val = 0;
+        for (int deg_a = 0; deg_a <= deg; deg_a++) {
+            T val_a = get(deg_a);
```
slaperche-scality: could be const?

src/polynomial.h
```diff
@@ -438,6 +440,57 @@ void Polynomial<T>::mul(Polynomial<T>* b)
...
+            if (val_a > 0) {
+                int deg_b = deg - deg_a;
+                T val_b = b->get(deg_b);
```
slaperche-scality: can be const?

src/polynomial.h
```diff
@@ -438,6 +440,57 @@ void Polynomial<T>::mul(Polynomial<T>* b)
...
+                    val = field->add(val, field->mul(val_a, val_b));
+            }
+        }
+        set(deg, val);
+    }
+    _degree = deg_out;
+}
+
+/** Multiply to polynomial (X + coef)
+ * \f{eqnarray*}{
```
slaperche-scality: Skip a line above, to have a separation between short desc and the rest

src/polynomial.h
```diff
@@ -438,6 +440,57 @@ void Polynomial<T>::mul(Polynomial<T>* b)
...
+ * P(X) = &a_0 + a_1 X + ... + a_n X^n \\
+ * Q(X) = &coef + X \\
+ * R(X) = &P(X) \times Q(X) \\
+ *      = &a_0 * coef + a_1 * coef X + ... + a_n X^n + \\
+ *        &a_0 X + a_1 X^2 + ... + a_{n-1} X^n + a_n X^{n+1} \\
+ *      = &a_0 * coef + sum_{i=1}^{n}(a_{i-1} + a_i * coef) X^i + a_n X^{n+1}
+ * \f}
+ *
+ * @param coef - coef of the multiplied polynomial (X + coef)
+ * @return
```
slaperche-scality: `@return` should be removed

src/polynomial.h
```diff
@@ -131,9 +131,9 @@ template <typename T>
 void Polynomial<T>::copy(Polynomial<T>* src)
```
slaperche-scality: Why do we have a copy function instead of overloading `operator=` or a copy constructor?

src/polynomial.h
```diff
@@ -131,9 +131,9 @@ template <typename T>
...
 {
     clear();
-
-    for (int i = src->degree(); i >= 0; i--)
-        set(i, src->get(i));
+    Term<T>* terms_src = src->get_terms();
+    terms.insert(terms_src->begin(), terms_src->end());
```
slaperche-scality: Could `terms` be non-empty here?

If so, it's not a copy but a concat/add.

src/polynomial.h
```diff
@@ -409,9 +409,8 @@ void Polynomial<T>::_derivative(Polynomial<T>* result, Polynomial<T>* a)
 template <typename T>
 void Polynomial<T>::neg()
 {
-    Polynomial<T> a(field), b(field);
-    b.copy(this);
-    _sub(this, &a, &b);
+    for (auto it = terms.begin(); it != terms.end(); it++)
```
slaperche-scality:
```c++

for (auto& term : terms) {
    terms[term.first] = field->neg(term.second);
}
```

Or even better: use `std::transform`

```c++

std::transform(terms.begin(), terms.end(), terms.begin(), A_LAMBDA_APPLYING_NEG);
```

src/polynomial.h
```diff
@@ -387,25 +387,6 @@ void Polynomial<T>::_extended_gcd(
     }
 }
 
-/** Derivative of a polynomial
```
slaperche-scality: commit message says: "improve derivative", commit message should also explain how

src/polynomial.h
```diff
@@ -506,12 +487,34 @@ void Polynomial<T>::mod(Polynomial<T>* b)
...
 template <typename T>
 void Polynomial<T>::derivative()
 {
-    Polynomial<T> a(field);
-    a.copy(this);
-    _derivative(this, &a);
+    auto it = terms.begin();
+    if (it->first == 0)
+        it++;
+    for (; it != terms.end(); it++) {
```
slaperche-scality: use `++it` instead of `it++`: this can avoid an useless copy of the iterator (if the compiler doesn't optimize it already).

src/fec_rs_fnt.h
```diff
@@ -90,12 +90,19 @@ class RsFnt : public FecCode<T> {
 
         int m = arith::get_smallest_power_of_2<int>(n_data);
         this->fft = new fft::Radix2<T>(this->gf, n, m, pkt_size);
+
+        // vector stores r^{-i} for i = 0, ... , k
+        T inv_r = this->gf->inv(r);
+        this->inv_r_powers = new vec::Vector<T>(this->gf, this->n_data + 1);
```
slaperche-scality: You could use a unique_ptr here, then you wouldn't need the delete in the destructor (it would be done automatically)

src/polynomial.h
vrancurel: agree

lamphamsy:
@all: the commits are updated to address comments. Thanks for your reviews :)

src/polynomial.h
```diff
@@ -97,19 +97,22 @@ class Polynomial {
     gf::Field<T>* field;
     T field_characteristic;
     Term terms;
+    int m_degree;
```
vrancurel: can you call it degree_cache to be explicit

src/fec_base.h
```diff
@@ -603,6 +626,126 @@ bool FecCode<T>::decode_bufs(
...
+    // vector x=(x_0, x_1, ..., x_k-1)
+    for (int i = 0; i < k; i++) {
+        vx->set(i, this->gf->exp(r, fragments_ids->get(i)));
+    }
+
+    for (int i = 0; i < k; i++) {
+        const int j = fragments_ids->get(i);
+        auto data = props[j].get(ValueLocation(offset, j));
+
+        if (data && *data == "@") {
```
vrancurel: the '@' deserves an explanation no ?

vrancurel: also please call the degree cache degree_cache

src/fec_rs_fnt.h
slaperche-scality: Why using a raw pointer here?

circle.yml
```diff
@@ -21,7 +21,7 @@ test:
     - ./lint.sh
     - cd build && make check-format
     - cd build && make check-lint
-    - cd build && make check
+    - cd build && env CTEST_OUTPUT_ON_FAILURE=1 make check
```
slaperche-scality: ho, nice one :+1: 

That being said, it may be better to do it in the CMakefile directly:

```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -183,7 +183,7 @@ coverage_evaluate()
 
 # Be sure to recompile the test driver if needed before running the test suite.
 add_custom_target(check
-  COMMAND ${CMAKE_CTEST_COMMAND}
+  COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure
   DEPENDS ${UNIT_TESTS}
   DEPENDS ${EC_DRIVER}
 )
```

src/fec_rs_gf2n_fft.h
```diff
@@ -71,11 +70,18 @@ class RsGf2nFft : public FecCode<T> {
         // std::cerr << "r=" << r << "\n";
 
         this->fft = new fft::CooleyTukey<T>(this->gf, n);
+
+        // vector stores r^{-i} for i = 0, ... , k
+        T inv_r = this->gf->inv(this->r);
+        this->inv_r_powers = new vec::Vector<T>(this->gf, this->n_data + 1);
```
slaperche-scality: Why using a raw pointer instead of` `unique_ptr`?

src/fec_rs_gfp_fft.h
```diff
@@ -120,11 +119,18 @@ class RsGfpFft : public FecCode<T> {
         } else {
             this->fft = new fft::CooleyTukey<T>(this->gf, n);
         }
+
+        // vector stores r^{-i} for i = 0, ... , k
+        T inv_r = this->gf->inv(this->r);
+        this->inv_r_powers = new vec::Vector<T>(this->gf, this->n_data + 1);
```
slaperche-scality: Why a raw pointer?

src/fec_base.h
```diff
@@ -207,7 +208,29 @@ class FecCode {
     }
 
   protected:
+    // primitive nth root of unity
+    T r;
```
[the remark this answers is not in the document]
lamphamsy: Yes, it's better. But this one will be in another PR :)

src/fec_base.h
```diff
@@ -207,7 +208,29 @@ class FecCode {
...
     gf::Field<T>* gf;
+    fft::FourierTransform<T>* fft = nullptr;
+    // This vector MUST be initialized by derived Class using multiplicative FFT
```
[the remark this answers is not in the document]
lamphamsy: Yes, it's correct.

CMakeLists.txt
```diff
@@ -119,6 +119,7 @@ endif()
 include(CheckCXXCompilerFlag)
 
 set(COMMON_CXX_FLAGS
+  -std=c++11
```
slaperche-scality: You don't need to do that, it's already done l. 38 by ` set(CMAKE_CXX_STANDARD 11)`

src/fec_base.h
```diff
@@ -603,6 +626,135 @@ bool FecCode<T>::decode_bufs(
...
+    vec::Vector<T>* output,
+    const std::vector<Properties>& props,
+    off_t offset,
+    vec::Vector<T>* fragments_ids,
+    vec::Vector<T>* words,
+    vec::Vector<T>* vx,
+    int vx_zero)
+{
+    if (inv_r_powers == nullptr) {
+        throw LogicError("The pointer \"inv_r_powers\" must be initialized");
```
slaperche-scality: very minor: you could use single quote ' or backtick `, that would sidestep the need to escape with \

src/fec_rs_fnt.h
```diff
@@ -83,18 +82,25 @@ class RsFnt : public FecCode<T> {
...
 
         // std::cerr << "n=" << n << "\n";
         // std::cerr << "r=" << r << "\n";
 
         int m = arith::get_smallest_power_of_2<int>(n_data);
         this->fft = new fft::Radix2<T>(this->gf, n, m, pkt_size);
+
+        // vector stores r^{-i} for i = 0, ... , k
+        T inv_r = this->gf->inv(this->r);
+        this->inv_r_powers = std::unique_ptr<vec::Vector<T>>(
```
slaperche-scality: I don't now if we want to do it, but if we bump from C++11 to C++14 (I did it for Quadiron, since C++14 is supported since several version for both GCC and Clang), you could use std::make_unique here.

Anyway, not to be done in this PR.

lamphamsy: yes, thanks. I noticed it for C++14 or later :)

lamphamsy:
The last change is to fix the check-format: do only reorder includes of fec_base.
So it's going to merge for next PR :)

PR #203, opened 2018-03-29
Decoding use polynomial extended on vec::Vector

lamphamsy:
1. It define Vec::Poly, a polynomial extended on vec::Vector
A vector of size n represents a polynomial P(X) of degree (n-1)
Vector: v[0] v[1] ... v[n-1] represents the polynomial:
P(X):= v[0] + v[1] X + ... + v[n-1] X^n

2. Implement Lagrange interpolation on vec::Poly


src/fec_rs_gf2n_fft_add.h
```diff
@@ -216,6 +215,82 @@ class RsGf2nFftAdd : public FecCode<T> {
             output->set(i, S.get(i));
     }
 
+    // Lagrange interpolation w/ vector
+    void decode_vec_Lagrange(
```
vrancurel: functions shall not contain capital letters

src/vec_poly.h
```diff
@@ -0,0 +1,234 @@
...
+}
+
+template <typename T>
+int Poly<T>::get_deg()
+{
+    return degree;
+}
+
+template <typename T>
+T Poly<T>::get(int exponent)
```
slaperche-scality: I think it would be better to overload the method [] instead of providing a method `get`

```
template <typename T>
const T& Poly<T>::operator[] (const int exponent) const
```

But this should be done in another PR, to replace all the `get` by` operator [] `in one sweep.

Maybe we can create an issue tagged improvement if we don't want to tackle it now (so we don't forget).

src/vec_poly.h
```diff
@@ -0,0 +1,234 @@
...
+#include "vec_vector.h"
+
+namespace nttec {
+namespace vec {
+
+/** A vector of size \f$n\f$ represents a polynomial \f$P(X)\f$ of degree
+ * \f$(n-1)\f$
+ *
+ * Vector: \f$v[0], v[1], ..., v[n-1]\f$ represents the polynomial:
+ *  \f$P(X):= v[0] + v[1] X + ... + v[n-1] X^n\f$
```
slaperche-scality: Wasn't that already the case for `nttec::Polynomial`? Is this class superseding `nttec::Polynomial` (lot of code seems duplicated otherwise, like `eval` is exactly the same code in both classes)?

I think I don't get the difference between `nettc::vec::Poly` and `nttec::Polynomial`: what's the difference? When do use one over the other?

src/vec_poly.h
```diff
@@ -0,0 +1,234 @@
...
+ * Note, derivate of \f$X^n\f$ is defined as sum of \f$n\f$ polynomials
+ * \f$X^{n-1}\f$, not a multiplication of \f$n\f$ and \f$X^{n-1}\f$. It can be
+ * expressed as below:
+ * \f{eqnarray*}{
+ *  derivative(X^n) &= (1 + 1 + ... + 1) X^{n-1}
+ *                  &= (n % p) * X^{n-1}
+ * \f{eqnarray*}{
+ * where \f$p\f$ is the characteristic of the field.
+ *
+ * @return
```
slaperche-scality: If it returns nothing, you can remove `@return` completely.

src/vec_poly.h
```diff
@@ -0,0 +1,234 @@
...
+void Poly<T>::zero()
+{
+    std::memset(buf, 0, this->n * sizeof(*buf));
+}
+
+template <typename T>
+void Poly<T>::dump()
+{
+    for (int i = 0; i <= degree; i++)
+        std::cout << buf[i] << " x^" << i << " + ";
```
slaperche-scality: This will result in a weird output, no? Somehting like "ax^0 + bx^1 + " (trailing +).

I think this could be fixed as follow:

```C++
    for (int i = 0; i <= degree; i++) {
        if (i) {
            std::cout << " + ";
        }
        std::cout << buf[i] << " x^" << i;
    }
```

src/fec_base.h
```diff
@@ -111,6 +112,10 @@ class FecCode {
     size_t pkt_size; // packet size, i.e. number of words per packet
     size_t buf_size; // packet size in bytes
 
+    // Length of operating codeword. It's calculated by derived Class
+    // FIXME: move n to protected
+    T n;
```
slaperche-scality: I'm not really fond of this kind of thing (we already have it for `inv_power_something`.

Relying on a child class to initialize a parameter of its parents is brittle and kind of break the encapsulation.

A better approach may be to have a abstract method (like `T codeword_len() = 0;`) that returns this info.
I see two benefit to this approach:
- you are sure that the child class will initialize it (i.e define the function) because it's a purely abstract method
- each child can compute how it want and don't need to know that it's parent has a field `n`.

src/fec_rs_fnt.h
```diff
@@ -79,16 +78,16 @@ class RsFnt : public FecCode<T> {
...
         // compute root of order n-1 such as r^(n-1) mod q == 1
-        this->r = this->gf->get_nth_root(n);
+        this->r = this->gf->get_nth_root(this->n);
 
         // std::cerr << "n=" << n << "\n";
         // std::cerr << "r=" << r << "\n";
 
         int m = arith::get_smallest_power_of_2<int>(n_data);
-        this->fft = new fft::Radix2<T>(this->gf, n, m, pkt_size);
+        this->fft = new fft::Radix2<T>(this->gf, this->n, m, pkt_size);
```
slaperche-scality: Not directly related to this patch, but I just noticed that we didn't document the param `n` in Doxygen for the `fft::Radix2` in `fft_2n.h`.

src/fec_rs_fnt.h
```diff
@@ -79,16 +78,16 @@ class RsFnt : public FecCode<T> {
         // with this encoder we cannot exactly satisfy users request, we need to
         // pad n = minimal divisor of (q-1) that is at least (n_parities +
         // n_data)
-        n = this->gf->get_code_len_high_compo(n_parities + n_data);
+        this->n = this->gf->get_code_len_high_compo(n_parities + n_data);
```
slaperche-scality: I don't think you need the `this->` prefix (same remarks for all the occurrence).

Also, `n` could have more explicit name I think (like you have `n_data`, this could be `n_something` (or something else, just not only `n`)).

src/fec_base.h
```diff
@@ -755,6 +769,75 @@ void FecCode<T>::decode_lagrange(
...
 
+template <typename T>
+void FecCode<T>::decode_vec_lagrange(
+    vec::Vector<T>* output,
+    const std::vector<Properties>& props,
+    off_t offset,
+    vec::Vector<T>* fragments_ids,
+    vec::Vector<T>* words,
+    vec::Vector<T>* vx,
+    int vx_zero)
```
slaperche-scality: Is `vx_zero` used?

src/fec_base.h
```diff
@@ -755,6 +769,75 @@ void FecCode<T>::decode_lagrange(
...
 }
 
+template <typename T>
+void FecCode<T>::decode_vec_lagrange(
+    vec::Vector<T>* output,
+    const std::vector<Properties>& props,
+    off_t offset,
+    vec::Vector<T>* fragments_ids,
+    vec::Vector<T>* words,
+    vec::Vector<T>* vx,
```
slaperche-scality: can be const? or even a const reference (if nullptr is not allowed).
Same remarks for `words` and `fragments_ids` above.

src/fec_base.h
```diff
@@ -755,6 +769,75 @@ void FecCode<T>::decode_lagrange(
...
+    vec::Poly<T> N_p_ifft(this->gf, n);
+    vec::Poly<T> S(this->gf, k);
+
+    // compute A(x) = prod_j(x-x_j)
+    A.zero();
+    A.set(0, 1);
+    for (int i = 0; i < k; i++) {
+        A.mul_to_x_plus_coef(this->gf->sub(0, vx->get(i)));
+    }
+    // std::cout << "A(x)="; A.dump();
```
slaperche-scality: These can be removed no? It's dead code (I guess it was useful when you were developping/debugging, but now it's unused).

src/fec_base.h
```diff
@@ -782,6 +783,9 @@ void FecCode<T>::decode_vec_lagrange(
     if (this->fft == nullptr) {
         throw LogicError("FEC base: FFT must be initialized");
     }
+    if (this->fft_full == nullptr) {
+        throw LogicError("FEC base: FFT must be initialized");
```
slaperche-scality: Same error message as above, but difference root cause: that can be misleading (could be "FEC base: FFT full must be initialized").

src/fec_rs_gf2n_fft_add.h
```diff
@@ -215,6 +215,85 @@ class RsGf2nFftAdd : public FecCode<T> {
...
+    void decode_vec_lagrange(
+        vec::Vector<T>* output,
+        const std::vector<Properties>& props,
+        off_t offset,
+        vec::Vector<T>* fragments_ids,
+        vec::Vector<T>* words,
+        vec::Vector<T>* vx,
+        int vx_zero)
+    {
+        // std::cout << "WORDS:"; words->dump();
```
slaperche-scality: These debug print can be removed, no?

src/fec_rs_gf2n_fft_add.h
```diff
@@ -215,6 +215,85 @@ class RsGf2nFftAdd : public FecCode<T> {
             output->set(i, S.get(i));
     }
 
+    // Lagrange interpolation w/ vector
+    void decode_vec_lagrange(
+        vec::Vector<T>* output,
+        const std::vector<Properties>& props,
+        off_t offset,
+        vec::Vector<T>* fragments_ids,
```
slaperche-scality: Can some of these (`fragments_ids`, `words` and `vx`) be const reference? Or are they all modified in this function?

src/fec_base.h
```diff
@@ -782,6 +783,9 @@ void FecCode<T>::decode_vec_lagrange(
     if (this->fft == nullptr) {
         throw LogicError("FEC base: FFT must be initialized");
     }
+    if (this->fft_full == nullptr) {
```
slaperche-scality: Hum, these checks begin to multiply a bit too much.

Maybe we should have another approach to be sure, by construction, that these case cannot happen.
I don't know if my suggestion about using an abstract accessor (in aother comment) is a good solution here as well, if not we can keep it like that for now, but I'll try to find a better way to handle that.

src/fec_rs_nf4.h
```diff
@@ -67,19 +66,26 @@ class RsNf4 : public FecCode<T> {
...
+        this->inv_r_powers = std::unique_ptr<vec::Vector<T>>(
+            new vec::Vector<T>(ngff4, this->n_data + 1));
+        for (int i = 0; i <= this->n_data; i++)
+            this->inv_r_powers->set(i, ngff4->exp(inv_r, i));
     }
 
     ~RsNf4()
     {
-        delete fft;
+        delete this->fft;
```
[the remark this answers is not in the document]
lamphamsy: `fft` was declared as a member of FEC::NF4: https://github.com/scality/nttec/pull/203/commits/e9613f8ec7caa66b2987978cf8f30cb4a570547b#diff-153495ed6caf85a68307a8491242e39cL145
I removed it since it's declared in the base class.
A separated commit will define fft and gf as a unique_ptr.

src/fec_rs_nf4.h
```diff
@@ -107,8 +113,8 @@ class RsNf4 : public FecCode<T> {
             words->set(i, ngff4->pack(words->get(i)));
         }
         // std::cout << "pack words:"; words->dump();
-        vec::ZeroExtended<T> vwords(words, n);
-        fft->fft(output, &vwords);
+        vec::ZeroExtended<T> vwords(words, this->n);
+        this->fft->fft(output, &vwords);
```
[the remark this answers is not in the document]
lamphamsy: because `fft` is now declared in the base class

src/fec_base.h
lamphamsy: yes, it's used in https://github.com/scality/nttec/pull/203/files#diff-827102cc0a2a7d910c0782ad775dd305R144

src/fec_rs_nf4.h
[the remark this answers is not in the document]
slaperche-scality: But child can access member of the base class without using `this` (if the visibility is correct, but the same apply for this): see https://ideone.com/LCqdw3

src/fec_base.h
slaperche-scality: ha, indeed.

src/vec_poly.h
lamphamsy: `nttec::Polynomial` is defined over std::map
`nttec::vec::Poly` is extended on vec::Vector, i.e. over an array
The two classes have same APIs that are used for decoding.

src/vec_poly.h
```diff
@@ -0,0 +1,234 @@
...
+    int degree;
+};
+
+template <typename T>
+Poly<T>::Poly(gf::Field<T>* field, int n) : Vector<T>(field, n)
+{
+    this->field = field;
+    this->field_characteristic = field->get_p();
+    this->buf = this->get_mem();
+    this->degree = 0;
```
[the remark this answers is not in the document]
lamphamsy: No, it's different. For example, a polynomial (X+1) can be defined over a vector of length 10. To operate with this polynomial, we want to iterate only the first two elements, not all 10 elements.

src/vec_poly.h
slaperche-scality: As seen with you, `nttec::Vec::Poly` is not really a vec (so it doesn't belong to the namespace `vec`), but it is a polynomial that uses a vector internally (like `nttec::Polynomial` is not a map, but a polynomial using a `std::map`

A more logical representation would be:
- an abstract base class `Polynomial` that defines the interface of a polynomial (eval, derive, …)
- a child class SparsePolynomial (that would correspond to the current `nttec::Polynomial`)
- a child class FullPolynomial (that would be your new `nttec::vec::Poly`).

Also, inheritance represents an IS-A relations, and you can't really say that a polynomial IS-A vector (nor IS-A map) so it would make more sense to use composition here (having a member Vec in your class).

As there require heavy change to your PR, let's forget it for now (but I'll create an issue to rework that).

src/fec_rs_nf4.h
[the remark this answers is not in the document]
slaperche-scality: Hum my bad, `this` is actually required when using template.

src/vec_poly.h
lamphamsy: I forgot that FFT will perform on `vec::Poly`. While input and output of FFT are vec::Vector (and vec::Buffer). Hence we should take into account this constraint.

src/fec_rs_gf2n_fft_add.h
lamphamsy: Yes, they could be. But these APIs will be changed in very next PRs. So I will pay attention at that time :)

src/fec_rs_nf4.h
```diff
@@ -48,39 +48,61 @@ namespace fec {
...
+            std::unique_ptr<fft::Radix2<T>>(new fft::Radix2<T>(ngff4, this->n));
     }
 
-    ~RsNf4()
+    inline void init_others()
     {
-        delete fft;
-        delete ngff4;
+        // vector stores r^{-i} for i = 0, ... , k
+        T inv_r = ngff4->inv(this->r);
```
slaperche-scality: can be const?

src/fec_rs_fnt.h
```diff
@@ -67,29 +66,52 @@ class RsFnt : public FecCode<T> {
...
-        // std::cerr << "n=" << n << "\n";
-        // std::cerr << "r=" << r << "\n";
+        int m = arith::get_smallest_power_of_2<int>(this->n_data);
+        this->fft = std::unique_ptr<fft::Radix2<T>>(
+            new fft::Radix2<T>(this->gf, this->n, m, this->pkt_size));
 
-        int m = arith::get_smallest_power_of_2<int>(n_data);
-        this->fft = new fft::Radix2<T>(this->gf, n, m, pkt_size);
+        this->fft_full = std::unique_ptr<fft::Radix2<T>>(
+            new fft::Radix2<T>(this->gf, this->n, this->n, this->pkt_size));
```
slaperche-scality: Not directly related to this patch, but I just noticed that we didn't document the param n in Doxygen for the fft::Radix2 in fft_2n.h.

src/fec_base.h
```diff
@@ -746,9 +778,72 @@ void FecCode<T>::decode_Lagrange(
...
+    vec::Vector<T>* fragments_ids,
+    vec::Vector<T>* words,
+    vec::Vector<T>* vx,
+    int vx_zero)
+{
+    if (this->fft == nullptr) {
+        throw LogicError("FEC base: FFT must be initialized");
+    }
+    if (this->fft_full == nullptr) {
+        throw LogicError("FEC base: FFT must be initialized");
```
slaperche-scality: Same error message as above, but difference root cause: that can be misleading (could be "FEC base: FFT full must be initialized").

src/fec_base.h
```diff
@@ -223,7 +248,16 @@ class FecCode {
...
 
-    virtual void decode_Lagrange(
+    virtual void decode_lagrange(
+        vec::Vector<T>* output,
+        const std::vector<Properties>& props,
+        off_t offset,
+        vec::Vector<T>* fragments_ids,
+        vec::Vector<T>* words,
+        vec::Vector<T>* vx,
+        int vx_zero = 0);
```
slaperche-scality: Do you need the default value here?

It seems that default value can have surprising behavior when used with virtual function and it's not recommanded to mix both. See https://google.github.io/styleguide/cppguide.html#Default_Arguments

If possible, we should avoid to mix virtual and default.

(same apply for `decode_prepare`).

PR #210, opened 2018-04-12
Ft/simd horizontal

lamphamsy:
### SIMD 128: add namespaces according to granularity

Granularity is type of integer that will be operated in parallel in operations of __m128i numbers.
It includes currently `u16` and `u32` for `uint16_t` and `uint32_t`

### Vector: shift two expensive operations to RingModN
Two expensive operations

- hadamard_mul(Doubled<T>* v)
- add(Doubled<T>* v)

that dominates FFT performance, are shifted to RingModN operations:

- hadamard_mul(int n, T* x, T* y)
- add(int n, T* x, T* y)

These operations would be accelerated by using vectorization.


src/fec_rs_gf2n.h
```diff
@@ -56,6 +56,7 @@ class RsGf2n : public FecCode<T> {
         RsMatrixType type)
         : FecCode<T>(FecType::SYSTEMATIC, word_size, n_data, n_parities)
     {
+        mat_type = type;
```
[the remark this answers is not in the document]
lamphamsy: The base class has member `type` indicating FEC type. So a different name is to avoid confusions

benchmark/benchmark.cpp
```diff
@@ -669,6 +669,15 @@ void run_scenario(Params_t* params)
     params->get_sizeof_T();
 
     switch (params->sizeof_T) {
+    case 2: {
```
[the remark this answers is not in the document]
lamphamsy: yes, it's better

src/fec_rs_gf2n.h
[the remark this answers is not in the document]
slaperche-scality: Ho, indeed.

Yes, it's better to rename it then.

benchmark/benchmark.cpp
```diff
@@ -663,48 +663,35 @@ void run(Benchmark<T>* bench, Params_t* params)
     }
 }
 
+template <typename T>
```
slaperche-scality: Better than my macro :+1:

src/simd_128_u16.h
```diff
@@ -0,0 +1,269 @@
...
+ *  It supports operations on 16-bit numbers
+ */
+namespace simd {
+namespace u16 {
+
+typedef __m128i am128i __attribute__((aligned(ALIGN_SIZE)));
+
+/* ==================== Essential Operations =================== */
+
+const aint16 F3 = 257;
```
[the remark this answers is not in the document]
slaperche-scality: I still think namespace are not the good way to do that.

Even without a proper SIMD wrapper (which we should write at some point), you can keep everything in the same namespace (`nttec::simd`), thanks to function overloading.

I mean, it's not a problem to define both:

```c++

inline aint128 mod_after_add(am128i a, aint32 card);
inline aint128 mod_after_add(am128i a, aint16 card);
```

They are different functions (because of their type signatures) and can have different body (using 16-bit and 32-bit element).
Did you tried? I think this could work until we get a proper wrapper.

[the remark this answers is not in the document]
lamphamsy: yeah, I will update it. I agree that it's better :)

lamphamsy:
@slaperche-scality : the code is updated. I removed u16 & u32 in simd namespace.

src/simd_128_u16.h
```diff
@@ -174,7 +164,7 @@ inline void mul_coef_to_buf(
 }
 
 inline void
-add_two_bufs(aint16* src, aint16* dest, size_t len, aint16 card = F3)
+add_two_bufs_u16(aint16* src, aint16* dest, size_t len, aint16 card = F3)
```
slaperche-scality: I think these `_u16` suffix are useless and can be removed (same comment for `mul_coef_to_buf`, `sub_two_bufs` and `mul_two_bufs` for both `_u16` and `_u32` suffix)

src/gf_ring.cpp
```diff
@@ -59,25 +59,25 @@ void RingModN<uint32_t>::sub_two_bufs(
...
     size_t len)
 {
-    simd::u32::sub_two_bufs(bufa, bufb, res, len, this->_card);
+    simd::sub_two_bufs_u32(bufa, bufb, res, len, this->_card);
 }
 
 template <>
 void RingModN<uint16_t>::add(int n, uint16_t* x_u16, uint16_t* y_u16)
 {
     int half = n / 2;
```
slaperche-scality: `half` in all of these functions can be declared const.

src/gf_ring.cpp
```diff
@@ -59,25 +59,25 @@ void RingModN<uint32_t>::sub_two_bufs(
...
     uint16_t* x_next = x_u16 + half;
-    simd::u16::add_two_bufs(y_u16, x_u16, half, this->_card);
-    simd::u16::add_two_bufs(y_u16, x_next, half, this->_card);
+    simd::add_two_bufs_u16(y_u16, x_u16, half, this->_card);
+    simd::add_two_bufs_u16(y_u16, x_next, half, this->_card);
 }
 
 template <>
 void RingModN<uint32_t>::add(int n, uint32_t* x_u32, uint32_t* y_u32)
```
slaperche-scality: The code for `add` and `hadamard_mul` is the same (after you remove the suffix) for both u16 and u32, I think it can be factorized by using `std::enable_if` + defining a trait such as `is_simd_accelerated`.

This can be done in another PR (when you'll add support for u64, to avoid having code triplicated).

src/simd_128_u16.h
lamphamsy: it's done

src/gf_ring.cpp
lamphamsy: it's done

src/gf_ring.cpp
lamphamsy: okay, I see. It will be optimized in other PR

PR #211, opened 2018-04-26
Vectorization on 256-bit registers

lamphamsy:
- Add SIMD vectorization on 256-bit registers supporting 16-bit and 32-bit integer granularity

- There are 3 possibilities for SIMD option:
  - OFF: no vectorization
  - SSE4: vector operations on 128-bit registers. It requires -msse4.1 flag
  - AVX2: vector operations on 256-bit registers. It requires -mavx2 flag


CMakeLists.txt
```diff
@@ -124,9 +133,12 @@ set(COMMON_CXX_FLAGS
 )
 
 # Option for enabling/disabling SIMD flags is for both of debug and release
-if (NTTEC_USE_SIMD)
+if (NTTEC_USE_SIMD STREQUAL "SSE4")
   SET(COMMON_CXX_FLAGS "${COMMON_CXX_FLAGS} -msse4.1")
-  add_definitions(-DNTTEC_USE_SIMD)
+  add_definitions(-DNTTEC_USE_SIMD=1)
+elseif (NTTEC_USE_SIMD STREQUAL "AVX2")
+  SET(COMMON_CXX_FLAGS "${COMMON_CXX_FLAGS} -mavx2 -msse4.1")
```
slaperche-scality: Is sse4.1 needed?

Activating AVX2 doesn't include SSE4.1 automatically?

circle.yml
```diff
@@ -24,5 +24,7 @@ test:
     - cd build && make check
     # w/o SIMD
     - cd build && cmake .. -DNTTEC_USE_SIMD=OFF && make benchmark
-    # w/ SIMD
-    - cd build && cmake .. -DNTTEC_USE_SIMD=ON && make benchmark
+    # w/ SIMD=SSE43
```
slaperche-scality: SSE43? Do you mean SSE 4.1?

src/simd_128_u32.h
```diff
@@ -190,8 +190,9 @@ inline void mul_coef_to_buf(
 
     m128i* _src = reinterpret_cast<m128i*>(src);
     m128i* _dest = reinterpret_cast<m128i*>(dest);
-    size_t _len = len / 4;
-    size_t _last_len = len - _len * 4;
+    unsigned ratio = sizeof(*_src) / sizeof(*src);
```
slaperche-scality: `ratio`, `_len` and `_last_len` can be const, no?

Same for the other occurences.

src/simd_nf4.h
```diff
@@ -0,0 +1,154 @@
...
+ */
+
+#ifndef __NTTEC_SIMD_NF4_H__
+#define __NTTEC_SIMD_NF4_H__
+
+#include <x86intrin.h>
+
+namespace nttec {
+
+/** The namespace simd contains functions for GF-NF4 that are accelerated by
```
slaperche-scality: I just noticed that we have 5 differents comments/documentation for the **simd** namespace (in simd_{128,256}_u{16,32}.h + 1 here)

I don't know how Doxygen will handle that, but it would be better to have a single one.
Something like

```
/** The namespace simd contains functions that are accelerated by using SIMD
 *  operations over 128 bits and 156 bits.
 *
 *  It supports operations on 16-bit and 32-bit numbers
 */
```

src/core.h
```diff
@@ -113,6 +113,53 @@ static inline std::mt19937& prng()
...
+#if NTTEC_USE_SIMD == SSE4
+#define ALIGN_SIZE 16
+#endif
+
+#if NTTEC_USE_SIMD == AVX2
+#define ALIGN_SIZE 32
+#endif
+
+template <typename T>
+inline T* aligned_allocate(size_t size)
```
slaperche-scality: Maybe we could use [std::align](http://en.cppreference.com/w/cpp/memory/align) to help here.

There is also [aligned_alloc](http://en.cppreference.com/w/c/memory/aligned_alloc), but it's only available in C++17 :slightly_frowning_face:

src/simd_256.h
```diff
@@ -42,6 +42,17 @@ const m256i F4minus1_m256i = _mm256_set1_epi32(65536); // NOLINT(cert-err58-cpp)
 const m256i F3_m256i = _mm256_set1_epi32(257);         // NOLINT(cert-err58-cpp)
 const m256i F3minus1_m256i = _mm256_set1_epi32(256);   // NOLINT(cert-err58-cpp)
 
+/* GCC doesn't include the split store intrinsics so define them here. */
+#if defined(__GNUC__) && !defined(__clang__) && !defined(__ICC)
```
slaperche-scality: If GNUC is defined, I don't think you need to check for clang and ICC, no?

src/simd_256.h
```diff
@@ -42,6 +42,17 @@ const m256i F4minus1_m256i = _mm256_set1_epi32(65536); // NOLINT(cert-err58-cpp)
...
+
+static inline void __attribute__((__always_inline__))
+_mm256_storeu2_m128i(__m128i* const hi, __m128i* const lo, const __m256i a)
+{
+    _mm_storeu_si128(lo, _mm256_castsi256_si128(a));
+    _mm_storeu_si128(hi, _mm256_extracti128_si256(a, 1));
+}
```
slaperche-scality: skip a line here (to have an empy line before the endif)

CMakeLists.txt
```diff
@@ -61,17 +61,24 @@ set(PACKAGE ${CMAKE_PROJECT_NAME})
...
+    CACHE
+    STRING
+    "Build with SIMD vectorizations."
+    FORCE
   )
 endif()
+set_property(CACHE NTTEC_USE_SIMD PROPERTY STRINGS
+  "OFF" "SSE4" "AVX2"
+)
 add_definitions(-DSSE4=1)
```
slaperche-scality: Why not defining these in the code directly?

Or even better: why not have "NTTEC_USE_AVX" and "NTTEC_USE_SSE" and set them accordingly, that way you can directly use `#if defined(NTTEC_USE_AVX)` instead of testing a numerical value.

e.g

```
# Option for enabling/disabling SIMD flags is for both of debug and release
if (NTTEC_USE_SIMD STREQUAL "SSE4")
  SET(COMMON_CXX_FLAGS "${COMMON_CXX_FLAGS} -msse4.1")
  add_definitions(-DNTTEC_USE_SSE)
elseif (NTTEC_USE_SIMD STREQUAL "AVX2")
  SET(COMMON_CXX_FLAGS "${COMMON_CXX_FLAGS} -mavx -msse4.2")
  add_definitions(-DNTTEC_USE_AVX)
endif()
```

Then, you don't need these definitions

src/simd.h
```diff
@@ -42,6 +42,12 @@ typedef uint32_t aint32 __attribute__((aligned(ALIGN_SIZE)));
 typedef uint64_t aint64 __attribute__((aligned(ALIGN_SIZE)));
 typedef __uint128_t aint128 __attribute__((aligned(ALIGN_SIZE)));
 
+// Decrease level of SIMD vectorization if it's not supported
+#if (NTTEC_USE_SIMD == AVX2) && not defined(__AVX2__)
+#undef NTTEC_USE_SIMD
+#define NTTEC_USE_SIMD SSE4
```
slaperche-scality: You could add a warning for the user

e.g:

```C
// Fallback to SEE if AVX is not supported.
#if defined(NTTEC_USE_AVX) && not defined(__AVX2__)
#warning "AVX enabled but not supported: fallback on SSE"
#undef NTTEC_USE_AVX
#define NTTEC_USE_SSE
#endif
```

CMakeLists.txt
lamphamsy: I'd like to keep this way since we could add vectorisations for AVX512

CMakeLists.txt
lamphamsy: When -msse4.1 is not here, there are errors: Illegal instruction. So I keep it like that. A further investigation could be useful to figure it out.

lamphamsy:
Aligned allocation is necessary to perform operations directly on __m128i and __m256i: we cast a pointer to pointer of __m128i/__m256i, e.g. https://github.com/scality/nttec/pull/211/files#diff-5d88a70921cd611d1bc150c69847cc17L169

src/simd_128_u16.h
```diff
@@ -36,9 +36,9 @@
 namespace nttec {
 
 /** The namespace simd contains functions that are accelerated by using SIMD
```
slaperche-scality: This comment should only be written once, not duplicated in every simd* files

src/simd_256.h
slaperche-scality: What about this one?

CMakeLists.txt
slaperche-scality: The problem with your approach is that there is nothing to enforce the synchronization between

```
add_definitions(-DSSE4=1)
add_definitions(-DAVX2=2)
```

And things like `-DNTTEC_USE_SIMD=2`.

If you changes the definitions to 3, maybe you'll keep using 2 here and that will be buggy: this approach is error prone.
You won't have this kind of issue with what I suggest.
I don't see the problem with supporting AVX512? In both case, you'll have to add another else branch.

.circleci/config.yml
```diff
@@ -5,11 +5,15 @@ jobs:
       - image: slaperche0scality/nttec:latest
     steps:
       - checkout
-      - run: mkdir build && cd build && cmake -G 'Unix Makefiles' .. -DCMAKE_BUILD_TYPE=Release  -DNTTEC_USE_SIMD=ON 'Unix Makefiles' .. && make
+      - run: mkdir build && cd build && cmake -G 'Unix Makefiles' .. -DCMAKE_BUILD_TYPE=Release 'Unix Makefiles' .. && make
       - run: cd build && make check-format
       - run: cd build && make check-lint
       - run: cd build && make check
       # w/ SIMD
       - run: cd build && make benchmark
```
slaperche-scality: This one is useless, no?
You already try every cases below.

simd.cmake
```diff
@@ -0,0 +1,91 @@
+#
+# Check if SSE4.1 or AVX2 instructions are available on the machine where the
+# project is compiled.
+# These commands are extracted and modified from
+# https://github.com/VcDevel/Vc/blob/master/cmake/FindSSE.cmake
+#
+if(CMAKE_SYSTEM_NAME MATCHES "Linux")
```
slaperche-scality: This file should be named `FindSSE.cmake` and be placed into `cmake/modules`.
Not here at the root of the project.

CMakeLists.txt
```diff
@@ -60,18 +60,15 @@ set(VERSION ${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}${VERSION_REV})
...
-    CACHE
-    STRING
-    "Build with SIMD vectorizations."
-    FORCE
-  )
-  set_property(CACHE NTTEC_USE_SIMD PROPERTY STRINGS
-    "OFF" "SSE4" "AVX2"
-  )
-endif()
+# Setting for SIMD level
```
slaperche-scality: About the commit message (since I can't comment directly on it…)

> If NTTEC_USE_SIMD is not given or given but not supported, default value
is used.

IMHO, if USE_SIMD is not defined, then SIMD should be disabled.
You should auto-detect the best SIMD instruction set only if USE_SIMD is enabled I think.

simd.cmake
slaperche-scality: Yeah, FindSIMD is good

src/simd_256.h
```diff
@@ -0,0 +1,60 @@
...
+
+// Disable `cert-err58-cpp` on these: AFAIK they cannot throw.
+// (probably a false positive present in Clang 5 and fixed in Clang 6).
+const m256i F4_m256i = _mm256_set1_epi32(65537);       // NOLINT(cert-err58-cpp)
+const m256i F4minus1_m256i = _mm256_set1_epi32(65536); // NOLINT(cert-err58-cpp)
+const m256i F3_m256i = _mm256_set1_epi32(257);         // NOLINT(cert-err58-cpp)
+const m256i F3minus1_m256i = _mm256_set1_epi32(256);   // NOLINT(cert-err58-cpp)
+
+/* GCC doesn't include the split store intrinsics so define them here. */
+#if defined(__GNUC__) && !defined(__clang__) && !defined(__ICC)
```
lamphamsy: @slaperche-scality : the following function is defined in clang or icc. If their define is removed, a duplication error occurs as clang defines GNUC also :(

slaperche-scality: Hum, indeed clang defines GNUC.

What about ICC? If we can't test it, we should remove it.

lamphamsy: yep :)

lamphamsy:
I squashed commits. I delete only useless multiple checks of NTTEC_USE_SIMD in simd.h:
```
diff --git a/src/simd.h b/src/simd.h
index 6948bcf..baf04ee 100644
--- a/src/simd.h
+++ b/src/simd.h
@@ -42,8 +42,6 @@ typedef uint32_t aint32 __attribute__((aligned(ALIGN_SIZE)));
 typedef uint64_t aint64 __attribute__((aligned(ALIGN_SIZE)));
 typedef __uint128_t aint128 __attribute__((aligned(ALIGN_SIZE)));

-#ifdef NTTEC_USE_SIMD
-
 namespace nttec {
 /** The namespace simd contains functions for GF-NF4 that are accelerated by
  *  using SIMD operations over 128bits
@@ -57,17 +55,13 @@ namespace simd {
 } // namespace simd
 } // namespace nttec

-#endif
-
 #ifdef NTTEC_USE_SSE4
 #include "simd_128.h"
 #elif defined NTTEC_USE_AVX2
 #include "simd_256.h"
 #endif

-#ifdef NTTEC_USE_SIMD
 #include "simd_nf4.h"
-#endif

 #endif // #ifdef NTTEC_USE_SIMD
```

PR #213, opened 2018-05-23
Introduce context for decoding

lamphamsy:
1. Decoding context
A context contains polynomials that are computed once, for a given set of indices of received fragments `fragments_ids = {x_1, x_2, .., x_k}`.
These polynomials will be served to decode original data given `k` received fragments.

Following polynomials are initialized by Context and does not change during the decoding:
- A := (x - x_1)(x - x_2)..(x-x_k)
- inv_A_i := 1 / (x_i * A_i(x_i))

Following polynomials are allocated by Context and will be used to docomputations during the decoding:
- N_p stores sum_i{n_i * x^z_i}
- N_p_ifft stores output of FFT(N_p)
- S stores decoded symbols

2. Refactor decoding into two steps
- Initialize context for given received fragments index
- Perform decoding using the computed context on received fragments


src/fec_base.h
```diff
@@ -165,10 +148,10 @@ class FecCode {
      *  if NON_SYSTEMATIC get_n_outputs()
      */
     virtual void decode(
+        ContextDec<T>* context,
```
vrancurel: Would prefer naming that DecodeContext

lamphamsy: It's renamed to DecodeContext :)

src/fec_base.h
```diff
@@ -222,31 +228,19 @@ class FecCode {
...
         const std::vector<Properties>& props,
         off_t offset,
-        vec::Vector<T>* fragments_ids,
-        vec::Vector<T>* words,
-        vec::Vector<T>* vx,
-        int vx_zero);
+        vec::Vector<T>* words);
 
-    virtual void decode_vec_lagrange(
+    virtual void decode_apply(
```
slaperche-scality: Can the `context` and `words` be null? If no, use a reference.

src/fec_base.h
```diff
@@ -576,52 +582,40 @@ bool FecCode<T>::decode_bufs(
     }
 
     vec::Vector<T> words(gf, n_words);
-    vec::Vector<T> fragments_ids(gf, n_words);
     vec::Vector<T> output(gf, n_data);
 
+    DecodeContext<T>* context = init_context_dec(&fragments_ids);
```
slaperche-scality: Should be a `unique_ptr`, then you don't need to delete l.637.

In fact, your code will leak memory if an exception is raised anywhere between here and your delete so you should really use an unique_ptr here, to have a memory handling that is exception-safe

src/fec_base.h
```diff
@@ -639,60 +633,137 @@ bool FecCode<T>::decode_bufs(
...
 }
 
 /**
  * Perform a Lagrange interpolation to find the coefficients of the
- * polynomial
+ * polynomial. The algorithm is mainly extracted from the paper \cite fnt-rs.
  *
  * @note If all fragments are available ifft(words) is enough
  *
+ * We have to find the first \f$k\f$ coefficients the numerator of the following
```
slaperche-scality: +1 for the doc :slightly_smiling_face:

src/fec_base.h
```diff
@@ -639,60 +633,137 @@ bool FecCode<T>::decode_bufs(
...
-    decode_vec_lagrange(
-        output, props, offset, fragments_ids, words, &vx, vx_zero);
+    decode_apply(context, output, words);
 }
 
-/* Prepare for decoding
+template <typename T>
+DecodeContext<T>* FecCode<T>::init_context_dec(vec::Vector<T>* fragments_ids)
+{
+    DecodeContext<T>* context =
```
slaperche-scality: You probably want a unique_ptr here

src/fec_base.h
```diff
@@ -147,10 +148,10 @@ class FecCode {
      *  if NON_SYSTEMATIC get_n_outputs()
      */
     virtual void decode(
+        DecodeContext<T>* context,
```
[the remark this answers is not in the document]
lamphamsy: yes, it's a nullptr in fec_rs_gf2n

src/fec_context.h
```diff
@@ -0,0 +1,262 @@
...
+        if (poly2_n)
+            delete poly2_n;
+    }
+
+    unsigned get_len_2k()
+    {
+        return len_2k;
+    }
+
+    vec::Poly<T>* get_A()
```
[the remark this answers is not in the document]
lamphamsy: only A_fft_2k and inv_A_i are not modified. Others will be modified during decoding

src/fec_context.h
```diff
@@ -0,0 +1,262 @@
...
+        // vector x=(x_0, x_1, ..., x_k-1)
+        for (int i = 0; i < k; ++i) {
+            if (betas->get(fragments_ids->get(i)) == 0) {
+                vx_zero = i;
+                break;
+            }
+        }
+    }
+
+    void dump()
```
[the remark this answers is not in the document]
lamphamsy: :)

src/gf_nf4.h
```diff
@@ -54,6 +54,7 @@ class NF4 : public gf::Field<T> {
     ~NF4();
     T card(void);
     T card_minus_one(void);
+    T get_inv_n_mod_p(int n);
```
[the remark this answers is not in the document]
lamphamsy: it's done. Other classes are updated also w/ `override`

src/gf_bin_ext.h
```diff
@@ -84,6 +84,9 @@ class BinExtension : public gf::Field<T> {
     void init_mask(void);
     void setup_tables(void);
     void setup_split_tables(void);
+    void hadamard_mul(int n, T* x, T* y);
+    void hadamard_mul_doubled(int n, T* x, T* y);
```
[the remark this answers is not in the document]
lamphamsy: I'd like to define GF's functions that perform directly operations on memory. Our defined classes, e.g. vec::Vector, vec::Doubled will use suitable GF's functions for their operations.
I keep `hadamard_mul_doubled`. And I rename `add(int n, T* x, T* y)` to `add_doubled(int n, T* x, T* y)` as you mentioned in an other comment.

src/gf_bin_ext.h
```diff
@@ -532,6 +535,48 @@ T BinExtension<T>::_inv_ext_gcd(T x)
...
+    }
+
+    // multiply y to the second half of `x`
+    for (int i = 0; i < half; i++) {
+        x_next[i] = mul(x_next[i], y[i]);
+    }
+}
+
+template <typename T>
+void BinExtension<T>::add(int n, T* x, T* y)
```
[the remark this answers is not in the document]
lamphamsy: it's done

src/fec_context.h
```diff
@@ -0,0 +1,262 @@
...
+#include <sys/time.h>
+
+#include "fft_base.h"
+#include "gf_base.h"
+#include "vec_poly.h"
+#include "vec_zero_ext.h"
+
+namespace nttec {
+
+/** Forward Error Correction code implementations. */
```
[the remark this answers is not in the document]
lamphamsy: it's done

src/fec_context.h
```diff
@@ -0,0 +1,262 @@
...
+    {
+        return poly2_2k;
+    }
+
+    vec::Poly<T>* get_S()
+    {
+        return S;
+    }
+
+    vec::Vector<T>* get_frag_ids()
```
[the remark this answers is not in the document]
lamphamsy: it's done

src/fec_context.h
```diff
@@ -0,0 +1,262 @@
...
+    {
+        return fragments_ids;
+    }
+
+    void set_frag_ids(vec::Vector<T>* fragments_ids)
+    {
+        this->fragments_ids = fragments_ids;
+    }
+
+    int get_vx_zero()
```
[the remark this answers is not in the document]
lamphamsy: it's done

src/fec_base.h
```diff
@@ -175,6 +176,8 @@ class FecCode {
         const std::vector<Properties>& input_parities_props,
         std::vector<std::ostream*> output_data_bufs);
 
+    virtual DecodeContext<T>* init_context_dec(vec::Vector<T>* fragments_ids);
```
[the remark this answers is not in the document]
lamphamsy: uhm, it should be public for our test: fec_utest uses this function to create context.

src/fec_base.h
lamphamsy: yes, it's done

src/fec_base.h
lamphamsy: :)

src/fec_base.h
lamphamsy: it's done

lamphamsy:
@slaperche-scality : the code is updated. Thanks for your reviews :)

src/vec_vector.h
```diff
@@ -369,6 +370,13 @@ bool Vector<T>::eq(Vector<T>* v)
     return true;
 }
 
+template <typename T>
+void Vector<T>::neg()
+{
+    for (int i = 0; i < this->n; i++)
```
slaperche-scality: You could use the new loop syntax:

```C++
for(auto &m : mem) {
    m = rn->neg(m);
}
```

src/vec_poly.h
```diff
@@ -145,6 +147,18 @@ void Poly<T>::derivative()
     degree--;
 }
 
+/** Derivative particularly for NF4 */
+template <typename T>
+void Poly<T>::derivative_nf4()
+{
+    for (int deg = 1; deg <= degree; ++deg) {
+        T _deg = field->replicate(deg % field_characteristic);
```
slaperche-scality: can be const

src/fec_context.h
```diff
@@ -40,13 +40,27 @@
...
+#include "gf_nf4.h"
 #include "vec_poly.h"
 #include "vec_zero_ext.h"
 
 namespace nttec {
 
 namespace fec {
 
+/* List of vectors handled by context */
+enum vector_type {
```
slaperche-scality: use `enum class`: they are strongly typed and the enumerators are scoped (unlike C-like enum).
Since they are scope, you could write:

```C++
enum class CtxVec {
    A = 0,
    A_FFT_2K,
    INV_A_I,
    S,
    N1,
    N2,
    2K_1,
    2K_2
};
```
And then use it as `CtxVec::A`, `CtxVec::2K_1`, …

Also, do you need `vec_end`? Seems unused to me.

src/fec_context.h
```diff
@@ -70,123 +84,98 @@ class DecodeContext {
...
 
-    void set_frag_ids(vec::Vector<T>* fragments_ids)
-    {
-        this->fragments_ids = fragments_ids;
-    }
-
-    int get_vx_zero()
-    {
-        return vx_zero;
+    vec::Vector<T>* get_vector(unsigned vec_type) const
```
slaperche-scality: If you use `enum class`, you can have `get_vector(CtxVec type)` since enum class are strongly type, the compiler will ensure that the value is not a random integer and thus you won't need a `default` case in your switch.

Since you no longer need to return `nullptr`, you can return a reference instead of a pointer (which is a good idea anyway: once an object is managed by a unique_ptr, you want avoid giving access to the raw pointer (because the user can then free the data, bypassing the unique_ptr safety) and prefer to use reference.

i.e, this could be written:

```C++
    vec::Vector<T>& get_vec(CtxVec type) const // or even simply vec instead of get_vec
    {
        switch (type) {
        case CtxVec::A_FFT_2K:
            return A_fft_2k.get();
        case CtxVec::INV_A_I:
            return inv_A_i.get();
        case CtxVec::VEC_N_1:
            return vec1_n.get();
        case CtxVec::N_2:
            return vec2_n.get();
        case CtxVec::2K_1:
            return vec1_2k.get();
        case CtxVec::2K_2:
            return vec2_2k.get();
        }
     }
```

src/fec_context.h
```diff
@@ -70,123 +84,98 @@ class DecodeContext {
...
+            return vec1_2k.get();
+        case vec_2k_2:
+            return vec2_2k.get();
+        default:
+            return nullptr;
+        }
     }
 
-    void set_vx_zero(int value)
+    vec::Poly<T>* get_poly(unsigned vec_type) const
```
slaperche-scality: If you have two set of "things" (vec and poly), then I think it would make sense to have two set of enum (CtxVec and CtxPoly)

Other than that, same remarks for get_vec.

src/fec_context.h
```diff
@@ -70,123 +84,98 @@ class DecodeContext {
...
+        else
+            A->set(0, 1);
         for (int i = 0; i < k; ++i) {
             A->mul_to_x_plus_coef(this->gf->sub(0, vx.get(i)));
         }
 
         // compute A'(x) since A_i(x_i) = A'_i(x_i)
         vec::Poly<T> _A(*A);
-        _A.derivative();
+        if (nf4)
```
slaperche-scality: Add braces {}, even for single lines (the linter don't enforce it yet, but as we enable more lint check this will probably becomes mandatory).
Ditto for the lines above (162-165)

src/fec_base.h
```diff
@@ -787,53 +788,53 @@ void FecCode<T>::decode_prepare(
...
+    unsigned len_2k = context.get_len_2k();
 
-    vec::Vector<T>* fragments_ids = context->fragments_ids;
-    std::shared_ptr<vec::Poly<T>> inv_A_i = context->inv_A_i;
-    std::shared_ptr<vec::Poly<T>> A_fft_2k = context->A_fft_2k;
-    std::shared_ptr<vec::Poly<T>> poly1_n = context->poly1_n;
-    std::shared_ptr<vec::Poly<T>> poly2_n = context->poly2_n;
-    std::shared_ptr<vec::Poly<T>> poly1_2k = context->poly1_2k;
-    std::shared_ptr<vec::Poly<T>> poly2_2k = context->poly2_2k;
+    vec::Vector<T>* fragments_ids = context.get_fragments_id();
```
slaperche-scality: I think these should be reference, because you shouldn't have a raw pointer from a `std::unique_ptr`

src/fec_rs_gf2n_fft_add.h
```diff
@@ -183,18 +186,18 @@ class RsGf2nFftAdd : public FecCode<T> {
...
+        const DecodeContext<T>& context,
         vec::Vector<T>* output,
         vec::Vector<T>* words) override
     {
-        vec::Vector<T>* fragments_ids = context->fragments_ids;
-        std::shared_ptr<vec::Poly<T>> A = context->A;
-        std::shared_ptr<vec::Poly<T>> inv_A_i = context->inv_A_i;
-        std::shared_ptr<vec::Poly<T>> poly1_n = context->poly1_n;
-        std::shared_ptr<vec::Poly<T>> S = context->S;
+        vec::Vector<T>* fragments_ids = context.get_fragments_id();
```
slaperche-scality: ditto, better if these are reference instead of pointers.

Also, the accessor could probably named `vec` and `poly`.

i.e
```C++
        vec::Poly<T>& A = context.poly(CtxPoly::A);
        vec::Vector<T>& inv_A_i = context.vector(CtxVec::INV_A_I);
        vec::Vector<T>& vec1_n = context.vec(CtxVec::N1);
        vec::Poly<T>& S = context.poly(CtxPoly::S);
```

src/fec_context.h
```diff
@@ -0,0 +1,262 @@
...
+    {
+        return vx_zero;
+    }
+
+    void set_vx_zero(int value)
+    {
+        this->vx_zero = value;
+    }
+
+    void init(vec::Vector<T>* fragments_ids, const vec::Vector<T>& vx)
```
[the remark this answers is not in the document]
slaperche-scality: I still don't understand the role of this method, compared to the constructor?

src/vec_doubled.h
```diff
@@ -57,7 +57,7 @@ class Doubled : public Vector<T> {
   public:
     explicit Doubled(Vector<T>* vec);
     const int get_n(void) const;
-    T get(int i);
+    T get(int i) const;
```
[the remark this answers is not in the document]
slaperche-scality: I think we're still returning a `T`, why?
Do we want to copy at each access?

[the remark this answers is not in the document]
lamphamsy: Yes, it's better. But there is a limit for vec_zero_ext: https://github.com/scality/nttec/blob/e107b430498e26782f16279b69354c3f6cca7c9e/src/vec_zero_ext.h#L84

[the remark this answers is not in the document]
slaperche-scality: Indeed, good catch.

We can keep it like that for now then

src/gf_base.h
```diff
@@ -74,9 +74,7 @@ Field<T>::Field(T p, int n, bool calculate_root)
 }
 
 template <typename T>
-Field<T>::~Field()
-{
-}
+Field<T>::~Field() = default;
```
slaperche-scality: minor remark: it could be declared inside the class directly, same for the others

src/fft_2.h
```diff
@@ -64,9 +64,7 @@ Size2<T>::Size2(const gf::Field<T>& gf) : FourierTransform<T>(gf, 2)
 }
 
 template <typename T>
-Size2<T>::~Size2()
-{
-}
+Size2<T>::~Size2() = default;
```
slaperche-scality: minor remark: it could be declared inside the class directly, same for the others

src/fec_context.h
```diff
@@ -140,17 +146,12 @@ class DecodeContext {
         }
     }
 
-    void init(
-        vec::Vector<T>& fragments_ids,
-        const vec::Vector<T>& vx,
-        bool nf4 = false)
+    void init(const vec::Vector<T>& vx)
```
slaperche-scality: Can this one be private?
I mean, what is it's use case (now that the init is fully handled by the constructor)?

src/gf_nf4.h
```diff
@@ -95,6 +95,7 @@ class NF4 : public gf::Field<T> {
 template <typename T>
 NF4<T>::NF4(unsigned n) : gf::Field<T>(T(65537), n)
 {
+    this->isNF4 = true;
```
slaperche-scality: Why do we need this?

slaperche-scality: Ok, I saw it.
I think it can be improved, but that's OK for now: you can ignore this comment.

lamphamsy:
I squashed commits only

PR #217, opened 2018-06-22
FFT::Radix2 uses bit-reversed permutation algorithm

lamphamsy:
Implement FFT Radix2 using bit-reversal permutation algorithm.

1. Decimation-in-time FFT
- Output is initialized as the input in the bit-reversal ordering
- Cooley-Tukey butterfly is performed on the output

It leads to an O(N*logN) complexity.

Note, we consider particular cases where the input is padded by zeros that always happens in erasure encoding process. Normally, FFT process starts performing butterfly operations on groups of size 2, then 4, 8 etc. Thanks to zero padding, we can start from a group size where there is at most a non-zero element of input in the group, all other elements of the group are zero. Indeed, after performing butterfly operation on the group, all elements equal to the non-zero one.

Given FFT length N, number of data length K, the output is initialized in the following clever way:
- Each group contains N/K elements
- Perform on groups containing a non-zero element of input: copy input's element at the bit-reversed index to all elements of the correspondent group of output
- For other groups, initialize them normally

It leads to an O(N*logK) complexity

2. Decimation-in-frequency FFT or inverse FFT
- Gentleman-Sande butterfly is performed on the input
- Output is copied from input at bit-reversed indices

It leads to an O(N*logN) complexity.

### DIFF
--- a/src/fft_2n.h
+++ b/src/fft_2n.h
@@ -169,33 +169,36 @@ void Radix2<T>::_init_bitrev()
 template <typename T>
 void Radix2<T>::fft(vec::Vector<T>& output, vec::Vector<T>& input)
 {
-    unsigned len = this->n;
+    const unsigned len = this->n;
+    const unsigned input_len = input.get_n();
+    // to support FFT on input vectors of length greater than from `data_len`
+    const unsigned group_len =
+        (input_len > data_len) ? len / input_len : len / data_len;
 
-    unsigned group_len = len / data_len;
-    unsigned idx = 0;
-    for (unsigned group = 0; group < data_len; ++group) {
+    for (unsigned idx = 0; idx < input_len; ++idx) {
         // set output  = scramble(input), i.e. bit reversal ordering
-        T a = input.get(rev[idx]);
-        output.set(idx, a);
-        unsigned end = idx + group_len;
-        for (unsigned i = idx + 1; i < end; ++i) {
+        const T a = input.get(idx);
+        const unsigned end = rev[idx] + group_len;
+        for (unsigned i = rev[idx]; i < end; ++i) {
             output.set(i, a);
         }
-        idx += group_len;
     }
-    // set output  = scramble(input), i.e. bit reversal ordering
-    for (; idx < len; idx++) {
-        output.set(idx, input.get(rev[idx]));
+    for (unsigned idx = input_len; idx < data_len; ++idx) {
+        // set output  = scramble(input), i.e. bit reversal ordering
+        const unsigned end = rev[idx] + group_len;
+        for (unsigned i = rev[idx]; i < end; ++i) {
+            output.set(i, 0);
+        }
     }
     // perform butterfly operations
     for (unsigned m = group_len; m < len; m *= 2) {
-        unsigned doubled_m = 2 * m;
-        unsigned ratio = len / doubled_m;
+        const unsigned doubled_m = 2 * m;
+        const unsigned ratio = len / doubled_m;
         for (unsigned j = 0; j < m; ++j) {
-            T r = W->get(j * ratio);
+            const T r = W->get(j * ratio);
             for (unsigned i = j; i < len; i += doubled_m) {
-                T a = output.get(i);
-                T b = this->gf->mul(r, output.get(i + m));
+                const T a = output.get(i);
+                const T b = this->gf->mul(r, output.get(i + m));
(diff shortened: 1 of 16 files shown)


src/gf_ring.h
```diff
@@ -459,6 +461,61 @@ inline void RingModN<T>::sub_vecp_to_vecp(
...
+ * \f{eqnarray*}{
+ *  P_i &= P_i + c \times Q_i \\
+ *  Q_i &= P_i - c \times Q_i \\
+ * \f}
+ *
+ * @param coef - coefficient, a finite field element
+ * @param buf1 - a buffer of `len` elements
+ * @param buf2 - a buffer of `len` elements
+ * @param len - number of elements per buffer
+ * @return
```
slaperche-scality: if it returns nothing, then remove the `@return`

src/gf_ring.h
```diff
@@ -459,6 +461,61 @@ inline void RingModN<T>::sub_vecp_to_vecp(
     }
 }
 
+/** Butterfly computation for Cooley-Tukey FFT algorithm
+ *
+ * Perform in-place oprations on two buffers `P`, `Q` with a coefficient `c`
+ * \f{eqnarray*}{
```
slaperche-scality: I didn't check the DOxygen output, but don't we need to skip a line before the `\f{eqnarray*}{`?

src/gf_ring.h
```diff
@@ -459,6 +461,61 @@ inline void RingModN<T>::sub_vecp_to_vecp(
...
+ *  P_i &= P_i + Q_i \\
+ *  Q_i &= c \times (P_i - Q_i)
+ * \f}
+ *
+ * @param coef - coefficient, a finite field element
+ * @param buf1 - a buffer of `len` elements
+ * @param buf2 - a buffer of `len` elements
+ * @param len - number of elements per buffer
+ * @return
```
slaperche-scality: ditto

src/gf_ring.h
```diff
@@ -459,6 +461,61 @@ inline void RingModN<T>::sub_vecp_to_vecp(
...
+        T b = mul(coef, buf2[i]);
+        buf1[i] = add(a, b);
+        buf2[i] = sub(a, b);
+    }
+}
+
+/** Butterfly computation for Gentleman-Sande FFT algorithm
+ *
+ * Perform in-place oprations on two buffers `P`, `Q` with a coefficient `c`
+ * \f{eqnarray*}{
```
slaperche-scality: ditto

src/gf_ring.cpp
```diff
@@ -58,6 +58,108 @@ void RingModN<uint32_t>::mul_coef_to_buf(
...
+        for (size_t i = vec_len * ratio; i < len; ++i) {
+            uint16_t a = buf1[i];
+            uint16_t b = mul(coef, buf2[i]);
+            buf1[i] = add(a, b);
+            buf2[i] = sub(a, b);
+        }
+    }
+}
+
+template <>
```
slaperche-scality: This is a copy paste of the 16-bit version…

I wonder if we could factorize this kind of code using a type trait (`is_vectorizable`) + `std::enable_if`

src/gf_ring.cpp
```diff
@@ -58,6 +58,108 @@ void RingModN<uint32_t>::mul_coef_to_buf(
...
+            uint16_t b = buf2[i];
+            uint16_t c = sub(a, b);
+            buf1[i] = add(a, b);
+            buf2[i] = mul(coef, c);
+        }
+    }
+}
+
+template <>
+void RingModN<uint32_t>::butterfly_gs(
```
slaperche-scality: ditto

src/gf_ring.cpp
slaperche-scality: I'll handle it in #222

src/fft_2n.h
```diff
@@ -39,44 +40,43 @@
 #include "vec_vector.h"
 #include "vec_zero_ext.h"
 
+/** Compute bit-reversed number for a given number
```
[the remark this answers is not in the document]
lamphamsy: we do agree that this change is not necessary as the gain is not significant and this computation is performed once

lamphamsy:
@slaperche-scality : thanks for your review. It's ready for next one :)

lamphamsy:

lamphamsy:
Oh, my bad. I forgot to change the base branch.

PR #218, opened 2018-06-22
FEC FNT: enhance encoding

lamphamsy:
For FEC over a prime field, the last step of the encoding process is to scan all symbols that are out of range of variable used for output. For example with FEC FNT(65537) with word_size = 2, i.e. variables are of uint16_t type, we scan symbols equal to (q-1) =2^16 that cannot be stored in uint16_t variable in output.

The step is refactored. If SIMD is enabled, it's accelerated using vector operations. Implementations are in the fec_vectorisation.cpp file.


lamphamsy:

PR #219, opened 2018-06-22
Simplify key/value property

lamphamsy:
A property carries extra-information (whose interpretation is left to the reader) related to a specific value (identified by its location).

It wraps a map whose each element is a key/value where
- key indicates the location of symbol whose value should be adjusted
- value is the value that could be used to adjust the symbol value

For prime fields, value is always 1.
For NF4, value is an uint32_t integer.

### DIFF
--- a/src/fec_base.h
+++ b/src/fec_base.h
@@ -787,14 +787,14 @@ void FecCode<T>::decode_prepare(
     const vec::Vector<T>& fragments_ids = context.get_fragments_id();
     for (unsigned i = 0; i < this->n_data; ++i) {
         const int j = fragments_ids.get(i);
-        auto data = props[j].get(ValueLocation(offset, j));
+        auto data = props[j].get(offset);
 
-        // Check if the symbol is a special case whick is marked by "@".
+        // Check if the symbol is a special case whick is marked by 1, i.e. true
         // Note: this check is necessary when word_size is not large enough to
         // cover all symbols of the field.
         // Following check is used for FFT over FNT where the single special
         // case symbol equals card - 1
-        if (data && *data == "@") {
+        if (data == 1) {
             words.set(i, this->gf->card() - 1);
         }
     }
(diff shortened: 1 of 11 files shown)


src/property.cpp
```diff
@@ -62,7 +50,7 @@ std::istream& operator>>(std::istream& is, Properties& props)
         }
         // Extract the key.
         auto end = line.find('=', begin);
-        std::string key = line.substr(begin, end - begin);
+        auto key = line.substr(begin, end - begin);
```
slaperche-scality: I think you can keep `std::string` here.
The line above use `auto` because the iterator type are very verbose, but for `std::string` I think it's clearer to keep it.

src/property.h
```diff
@@ -40,91 +40,41 @@
...
 class Properties {
   public:
-    inline void add(const ValueLocation& loc, const std::string& data)
+    inline void add(const off_t loc, const uint32_t data)
     {
         props[loc] = data;
     }
 
-    inline const std::string* const get(const ValueLocation& loc) const
+    inline const uint32_t get(const off_t loc) const
```
slaperche-scality: I think you can remove the const here, since you return by value.

src/fec_base.h
```diff
@@ -796,14 +796,14 @@ void FecCode<T>::decode_prepare(
...
+        auto data = props[j].get(offset);
 
-        // Check if the symbol is a special case whick is marked by "@".
+        // Check if the symbol is a special case whick is marked by 1, i.e. true
         // Note: this check is necessary when word_size is not large enough to
         // cover all symbols of the field.
         // Following check is used for FFT over FNT where the single special
         // case symbol equals card - 1
-        if (data && *data == "@") {
+        if (data == 1) {
```
slaperche-scality: Could we use a define/global constant instead of `1`.
Having a name would be more readable and explicit.

src/property.cpp
lamphamsy: it's done

src/property.h
lamphamsy: it's done

src/fec_base.h
lamphamsy: it's done. Thanks

lamphamsy:

PR #220, opened 2018-07-26
Ft/rename nttec to quadiron

slaperche-scality:


slaperche-scality:
`quadiron` is a bit long I think.
In QuadIron (now know as DCSS) we were using `kad` (because it was based on Kademlia), so here I used `quad`.

I'm open to better suggestion :)

PR #221, opened 2018-07-27
rename the namespace to `quadiron`

slaperche-scality:
`quad` is a bit too "common" to be used as a good namespace.
Using `quadiron` seems better to avoid eventual name clashes, and if
it's too long users can still shorten it by themselves.


slaperche-scality:
@lamphamsy in the end, I think you were right: `quadiron` is better as namespace.

PR #227, opened 2018-08-03
Remove GFP and Polynomial

slaperche-scality:
They're not necessary for QuadIron


lamphamsy:
@slaperche-scality : I understand that polynomial and GF::Extension could be removed but why is FEC over GFP removed? In the title, it's GF::Extension, not GFP (prime fields)

slaperche-scality:
@lamphamsy the issue was named "Remove gpfn and polynomial.h" so I though that "gfpn" was GFP, was that wrong?

lamphamsy:
@slaperche-scality : yes, GFP and GFPN are different

slaperche-scality:
@lamphamsy ho, then we should probably revert the last commit of this PR.

What code should be removed to remove GFPN?

lamphamsy:
@slaperche-scality : yes, we should revert it. We have not RS codes for GFPN, I think we have only the GFPN class and its utest.

Issue #160, opened 2018-02-09
cleanup ntl.h


slaperche-scality:
ntl.h is the "public" header that will be used by the users of the library, we should cleanup the included headers to be sure we only expose "private" API and not internals of the library

slaperche-scality:
Indeed, I'll update the issue

PR #228, opened 2018-08-03
Define a public API

slaperche-scality:
`quadiron.h` is the public header, as such it should only
contains/expose the public interface of the library (and not everything,
as that was the case).

Refs: #160

### DIFF
--- a/src/core.h
+++ b/src/core.h
@@ -96,15 +96,6 @@ struct GroupedValues {
     uint32_t flag;
 };
 
-/** Return the version string of QuadIron.
- *
- * The version string has the form MAJOR.MINOR.PATCH-REVISION, where '-REVISION'
- * is optional (only present for development version).
- *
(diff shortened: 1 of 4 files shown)


slaperche-scality:
WIP: I still need to write a short example to illustrate how to use the API (and to ensure that what is exported by `quadiron.h` is enough to use the library).

slaperche-scality:
Ready to merge once I get another approval.

The example will be done in another PR.

src/gf_ext.h
```diff
@@ -33,6 +33,7 @@
 
 #include "arith.h"
 #include "gf_base.h"
+#include "polynomial.h"
```
lamphamsy: This file would be removed?

slaperche-scality: Yes, I'll do it during the rebase

PR #232, opened 2018-08-10
Use GoogleTest for our unit tests

slaperche-scality:
Port our existing test suite on GoogleTest

### DIFF
--- a/src/gf_ring.h
+++ b/src/gf_ring.h
@@ -107,7 +107,7 @@ class RingModN {
         T x,
         T h,
         std::vector<T>* primes,
-        std::vector<int>* exponent) const;
+        std::vector<int>* exponents) const;
     T get_order(T x) const;
     virtual T get_nth_root(T n) const;
     T get_code_len(T n) const;
@@ -121,10 +121,10 @@ class RingModN {
   protected:
     T _card;
     T root;
-    std::unique_ptr<std::vector<T>> primes = nullptr;
-    std::unique_ptr<std::vector<int>> exponent = nullptr;
-    std::unique_ptr<std::vector<T>> all_primes_factors = nullptr;
-    std::unique_ptr<std::vector<T>> proper_divisors = nullptr;
+    std::vector<T> primes;
+    std::vector<int> exponents;
+    std::vector<T> all_primes_factors;
+    std::vector<T> proper_divisors;
 };
 /**
  * The calculation of primitive roots is performed in the constructor.
@@ -141,13 +141,6 @@ RingModN<T>::RingModN(T card, bool calculate_root)
 {
     this->_card = card;
     this->root = 0;
-
-    this->primes = std::unique_ptr<std::vector<T>>(new std::vector<T>());
-    this->exponent = std::unique_ptr<std::vector<int>>(new std::vector<int>());
-    this->all_primes_factors =
-        std::unique_ptr<std::vector<T>>(new std::vector<T>());
-    this->proper_divisors =
-        std::unique_ptr<std::vector<T>>(new std::vector<T>());
 
     // Init: computing primitive root
     // NOTE: for a derived classe, it should perform this operation in its
@@ -458,15 +451,14 @@ void RingModN<T>::compute_factors_of_order()
 {
     T h = RingModN<T>::card_minus_one();
     // prime factorisation of order, i.e. order = p_i^e_i where
-    //  p_i, e_i are ith element of this->primes and this->exponent
-    arith::factor_prime<T>(h, primes.get(), exponent.get());
-    // store all primes in a vector. A prime is replicated according to its
-    //  exponent
-    arith::get_prime_factors_final<T>(
-        primes.get(), exponent.get(), all_primes_factors.get());
+    //  p_i, e_i are ith element of this->primes and this->exponents.
+    arith::factor_prime<T>(h, &primes, &exponents);
+    // Store all primes in a vector.
+    // A prime is replicated according to its exponent.
+    all_primes_factors = arith::get_prime_factors<T>(primes, exponents);
     // calculate all proper divisor of order. A proper divisor = order/p_i for
     //  each prime divisor of order.
-    arith::get_proper_divisors<T>(h, primes.get(), proper_divisors.get());
+    proper_divisors = arith::get_proper_divisors<T>(h, primes);
 }
 
 /**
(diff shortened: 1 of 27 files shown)


src/gf_ring.h
```diff
@@ -324,10 +324,10 @@ T RingModN<T>::exp_quick(T base, T exponent) const
 
     T tmp = this->exp_quick(base, exponent / 2);
     // std::cout << base << "^" << exponent / 2 << "=" << tmp << "\n";
-    result = this->mul(tmp, tmp);
+    result = RingModN<T>::mul(tmp, tmp);
```
slaperche-scality: This look like a bad fix, it break the tests…

test/buffers_utest.cpp
```diff
@@ -63,222 +55,156 @@ class BuffersUtest {
...
+using TestedTypes = ::testing::Types<uint32_t, uint64_t>;
+TYPED_TEST_CASE(BuffersTest, TestedTypes);
 
-        std::vector<T*> mem3(end - begin, nullptr);
-        for (int i = 0; i < end - begin; i++) {
-            mem3[i] = mem1->at(i + begin);
-        }
-        quadiron::vec::Buffers<T> vec3(end - begin, size, &mem3);
-        // vec3.dump();
+TYPED_TEST(BuffersTest, Test1) // NOLINT
```
slaperche-scality: @lamphamsy If you have a better name than Test1, I'll be glad to rename it (it was named utest1 before so I kept it like that)

test/buffers_utest.cpp
```diff
@@ -63,222 +55,156 @@ class BuffersUtest {
...
-        std::cout << "buffers_utest4\n";
-        int i;
-        int word_size;
+    ASSERT_EQ(vec2, vec3);
+}
 
-        for (i = 0; i <= quadiron::arith::log2<T>(sizeof(T)); i++) {
-            word_size = quadiron::arith::exp2<T>(i);
-            pack_unpack(word_size);
+TYPED_TEST(BuffersTest, Test2) // NOLINT
```
slaperche-scality: @lamphamsy If you have a better name than Test2, I'll be glad to rename it (it was named utest2 before so I kept it like that)

test/buffers_utest.cpp
```diff
@@ -63,222 +55,156 @@ class BuffersUtest {
...
+}
 
-    void pack_unpack(int word_size)
-    {
-        std::cout << "pack_unpack with word_size=" << word_size << "\n";
-        int n = 8;
-        int size = 32;
-        int bytes_size = size * word_size;
-        int i, j, t, u;
+TYPED_TEST(BuffersTest, Test3) // NOLINT
```
slaperche-scality: @lamphamsy If you have a better name than Test3, I'll be glad to rename it (it was named utest3 before so I kept it like that)

test/mat_utest.cpp
```diff
@@ -27,14 +27,17 @@
...
 
-void mat_utest1()
-{
-    std::cout << "mat_utest1\n";
+namespace gf = quadiron::gf;
+namespace vec = quadiron::vec;
 
-    quadiron::gf::Prime<uint32_t> gf11(11);
-    quadiron::vec::Matrix<uint32_t> mat(gf11, 3, 3);
+TEST(MatrixTest, Test1) // NOLINT
```
slaperche-scality: @lamphamsy If you have a better name than Test1, I'll be glad to rename it (it was named utest1 before so I kept it like that)

test/mat_utest.cpp
```diff
@@ -48,23 +51,21 @@ void mat_utest1()
...
+    ASSERT_EQ(mat.get(1, 0), 7);
+    ASSERT_EQ(mat.get(1, 1), 9);
+    ASSERT_EQ(mat.get(1, 2), 10);
+    ASSERT_EQ(mat.get(2, 0), 0);
+    ASSERT_EQ(mat.get(2, 1), 6);
+    ASSERT_EQ(mat.get(2, 2), 5);
 }
 
-void mat_utest2()
+TEST(MatrixTest, Test2) // NOLINT
```
slaperche-scality: @lamphamsy If you have a better name than Test2, I'll be glad to rename it (it was named utest2 before so I kept it like that)

test/fec_utest.cpp
```diff
@@ -27,156 +27,119 @@
...
         quadiron::vec::Vector<T> v_p(*gf, n_data);
         std::vector<int> ids;
-        for (int i = 0; i < code_len; i++)
+
+        for (int i = 0; i < code_len; i++) {
             ids.push_back(i);
+        }
+
         std::vector<quadiron::Properties> props(code_len);
         for (int j = 0; j < 1000; j++) {
```
slaperche-scality: @lamphamsy why are we iterating 1000 times?

It's the case in several tests and it makes some tests slow, is there a reason why we are doing this?

src/gf_ring.h
slaperche-scality: @lamphamsy any idea on how to fix this?

On one hand it seems that we must call the child's implementation, but on the other hand calling a virtual function in a constructor is really really error-prone (because the object is not fully initialized).

test/fec_utest.cpp
lamphamsy: Generally, we should test for different input data. Here, we test for different input data `v` and received fragments lists

slaperche-scality: Ok, I see.

That being said, why 1000? Is there some statistical reason behind it? Or would 100 be just as fine?

lamphamsy: Personally I think 100 is a bit small. But if the running time for 1000 is too long, we could choose a smaller one.

test/buffers_utest.cpp
lamphamsy: It could be renamed as: `Check constructors`

test/fec_utest.cpp
slaperche-scality: Ok, we can keep 1000 for now and see if it's a problem later (as we add more tests)

test/buffers_utest.cpp
lamphamsy: It could be renamed as: `Even odd separation`

test/buffers_utest.cpp
lamphamsy: Here is:  `Zero-extended constructor checking`

test/buffers_utest.cpp
slaperche-scality: Done

test/buffers_utest.cpp
slaperche-scality: Done

test/buffers_utest.cpp
slaperche-scality: Done

test/mat_utest.cpp
slaperche-scality: Done

test/mat_utest.cpp
slaperche-scality: Done

PR #236, opened 2018-08-31
Decoding enhancement

lamphamsy:
### Enhance inverse FFT2n

iFFT does two not-so-good things on input:
- shuffle output's elements according to an reversed bit order.
- modify input vector/buffer

These actions will be avoided by swapping elements on output. Note, input is in reversed-bit order. Hence butterfly operations can be performed directly on this order. But the output must be re-order
in the natural order by using swap operations.
To preserve output's order, we shuffle the output twice, before and after the butterfly operation. 
A swap operation is to swap only two buffers' pointer, hence it's not costly.

### Reduce auxilier buffers in context

For decoding using Buffers, context allocates buffers of total length `(k + max(n, 2k))` instead of `(2n + 3k` previously.

### Simplify decoding

A step doing zero-out a Buffers is avoided.


src/vec_buffers.h
```diff
@@ -118,6 +118,7 @@ class Buffers {
     void separate_even_odd(Buffers<T>& even, Buffers<T>& odd);
     friend bool operator==<T>(const Buffers<T>& lhs, const Buffers<T>& rhs);
     void dump(void);
+    void swap(unsigned i, unsigned j);
```
slaperche-scality: Could be `size_t` instead of `unsigned`.

src/vec_buffers.h
```diff
@@ -405,6 +406,12 @@ bool operator==(const Buffers<T>& lhs, const Buffers<T>& rhs)
     return true;
 }
 
+template <typename T>
+void Buffers<T>::swap(unsigned i, unsigned j)
+{
+    std::swap(mem[i], mem[j]);
```
slaperche-scality: Unlikely to matter in our specific case, but it's considered a good practice to write it as:

```C++
void Buffers<T>::swap(unsigned i, unsigned j)
{
    using std::swap;
    swap(mem[i], mem[j]);
}
```

That way, [ADL](https://en.wikipedia.org/wiki/Argument-dependent_name_lookup) can kick in and if a specialized version of `swap` exists it will be used.

src/vec_vector.h
```diff
@@ -411,7 +413,19 @@ void Vector<T>::to_poly(Polynomial<T>* poly)
...
-void Vector<T>::dump(void)
+void Vector<T>::sort()
+{
+    std::sort(mem, mem + n - 1);
+}
+
+template <typename T>
+void Vector<T>::swap(unsigned i, unsigned j)
+{
+    std::swap(mem[i], mem[j]);
```
slaperche-scality: Same remark as in the previous commit.

src/fft_2n.h
```diff
@@ -207,6 +207,7 @@ void Radix2<T>::fft(vec::Vector<T>& output, vec::Vector<T>& input)
 }
 
 /** Perform decimation-in-frequency FFT or inverse FFT
+ * Note: input vector is in reversed-bit order
```
slaperche-scality: Skip a line and use Doxygen syntax (`@note`, or even `@pre` if it's a pre-condition).

src/fft_2n.h
```diff
@@ -219,23 +220,29 @@ void Radix2<T>::fft(vec::Vector<T>& output, vec::Vector<T>& input)
 template <typename T>
 void Radix2<T>::fft_inv(vec::Vector<T>& output, vec::Vector<T>& input)
 {
-    unsigned len = this->n;
+    const unsigned len = this->n;
+
+    // copy input to output
```
slaperche-scality: Useless comment.

src/fft_2n.h
```diff
@@ -219,23 +220,29 @@ void Radix2<T>::fft(vec::Vector<T>& output, vec::Vector<T>& input)
...
     for (unsigned m = len / 2; m >= 1; m /= 2) {
         unsigned doubled_m = 2 * m;
         for (unsigned j = 0; j < m; ++j) {
             T r = inv_W->get(j * len / doubled_m);
             for (unsigned i = j; i < len; i += doubled_m) {
-                T a = input.get(i);
-                T b = input.get(i + m);
-                input.set(i, this->gf->add(a, b));
-                input.set(i + m, this->gf->mul(r, this->gf->sub(a, b)));
+                T a = output.get(i);
```
slaperche-scality: `a` and `b` can be `const`, no?

src/fft_2n.h
```diff
@@ -219,23 +220,29 @@ void Radix2<T>::fft(vec::Vector<T>& output, vec::Vector<T>& input)
...
+                T b = output.get(i + m);
+                output.set(i, this->gf->add(a, b));
+                output.set(i + m, this->gf->mul(r, this->gf->sub(a, b)));
             }
         }
     }
 
-    // set output  = scramble(input), i.e. bit reversal ordering
+    // reversion of elements of output to return values on the natural order
```
slaperche-scality: You could have a method `reverse`, it would be clearer (and avoid code duplication).

src/fft_2n.h
```diff
@@ -297,33 +304,51 @@ void Radix2<T>::fft(vec::Buffers<T>& output, vec::Buffers<T>& input)
...
+    const unsigned len = this->n;
+
+    // 1st reversion of elements of output
+    for (unsigned i = 0; i < len; i++) {
+        if (rev[i] < i) {
+            output.swap(i, rev[i]);
+        }
+    }
+
+    // copy input to output
```
slaperche-scality: useless comment

src/fft_2n.h
```diff
@@ -297,33 +304,51 @@ void Radix2<T>::fft(vec::Buffers<T>& output, vec::Buffers<T>& input)
...
 
-    unsigned len = this->n;
     for (unsigned m = len / 2; m >= 1; m /= 2) {
         unsigned doubled_m = 2 * m;
         for (unsigned j = 0; j < m; ++j) {
             T r = inv_W->get(j * len / doubled_m);
             for (unsigned i = j; i < len; i += doubled_m) {
-                T* a = input.get(i);
-                T* b = input.get(i + m);
+                T* a = output.get(i);
```
slaperche-scality: Can `a` and `b` be `const`?

src/fft_2n.h
```diff
@@ -297,33 +304,51 @@ void Radix2<T>::fft(vec::Buffers<T>& output, vec::Buffers<T>& input)
 }
 
 /** Perform decimation-in-frequency FFT or inverse FFT
+ * Input buffer is in reversed-bit order. Hence butterfly operations can
```
slaperche-scality: Skip a line before the lengthy description (that way Doxygen can format things properly)

src/fft_2n.h
```diff
@@ -297,33 +304,51 @@ void Radix2<T>::fft(vec::Buffers<T>& output, vec::Buffers<T>& input)
...
+ * be performed directly on this order. But the output must be re-order in the
+ * natural order by using swap operations. A swap operation is to swap only two
+ * buffers' pointer, hence it's not costly. However, it will modify the output
+ * buffers. To avoid that, we shuffle the output twice, before and after the
```
slaperche-scality: It's not really a shuffle, you just reverse it. (shuffling is usually used in term of random shuffling).

src/vec_buffers.h
```diff
@@ -261,6 +271,58 @@ Buffers<T>::Buffers(const Buffers<T>& vec1, const Buffers<T>& vec2)
     mem.insert(mem.end(), vec2.get_mem().begin(), vec2.get_mem().end());
 }
 
+/**
```
slaperche-scality: You should add some unit tests for this new constructor.

src/vec_buffers.h
```diff
@@ -261,6 +271,58 @@ Buffers<T>::Buffers(const Buffers<T>& vec1, const Buffers<T>& vec2)
...
+ * Constructor of Buffers whose elements are shuffled from a given Buffers.
+ * An extending is necessary if the output length is greater than that given
+ * Buffers' length.
+ *
+ * Ex: input [p1, p2, p3, p4, p5] and map [3, 6, 1, 5, 4, 0, 2]
+ * Output: [p3, 0, p1, 0, p4, p0, p2] where `0` is an all-zero buffer.
+ *
+ * @param vec - a given Buffers instance of `m` elements
+ * @param map - a vector of `k` elements
```
slaperche-scality: the parameter `n` is not documented

src/vec_buffers.h
```diff
@@ -255,27 +255,23 @@ Buffers<T>::Buffers(const Buffers<T>& vec1, const Buffers<T>& vec2)
...
             for (int i = 0; i < n; i++) {
-                aligned_deallocate<T>(this->mem->at(i));
+                aligned_deallocate<T>(mem[i]);
             }
         } else if (this->mem_alloc_case == BufMemAlloc::ZERO_EXTEND) {
             aligned_deallocate<T>(this->zeros);
         }
-        this->mem->shrink_to_fit();
-        delete this->mem;
+        mem.shrink_to_fit();
```
slaperche-scality: Why calling `shrink_to_fit` here? (this may cause a reallocation and doesn't seems really useful here)

src/fec_rs_fnt.h
```diff
@@ -35,10 +35,8 @@
 #include "fec_base.h"
 #include "fft_2n.h"
 #include "gf_prime.h"
-#include "vec_buf_zero_ext.h"
```
slaperche-scality: Yeah less code \o/ (less bug)

src/vec_buffers.h
```diff
@@ -101,23 +101,23 @@ class Buffers {
     Buffers(const Buffers<T>& vec, int n = 0);
     Buffers(const Buffers<T>& vec, int begin, int end);
     Buffers(const Buffers<T>& vec1, const Buffers<T>& vec2);
-    virtual ~Buffers();
-    virtual int get_n(void) const;
-    virtual size_t get_size(void) const;
+    ~Buffers();
```
slaperche-scality: This is safe because nobody inherits from this class anymore, but if someone want to inherit from it it must remember to make the destructor virtual again and this can be easily overlooked.

To be extra-safe, you should declare the class as `class Buffers final {` at l.97

That way, you cannot inherit from `Buffers` (and if you really want to, you need to come edit the class to remove the `final` and if you do that, you are more likely to also edit the destructor to make it virtual again).

src/vec_buffers.h
lamphamsy: `i` and `j` are for indices of buffers which is used as `int`. So I think `unsigned` is better.

slaperche-scality: If they are indices, we should use `size_t`: that's the purpose of the type ("size_t is commonly used for array indexing and loop counting." from [here](https://en.cppreference.com/w/c/types/size_t) for instance)

If indices are used as `int` (but why would they: a negative index doesn't make sense to me) we shouldn't pass an `unsigned` type: mixing signed and unsigned is asking for trouble because it's easy to get subtle bug (thanks to the weird promotion rules of C and C++).

lamphamsy: yes, I see that. I will synchronize all indices in another commit.

test/buffers_utest.cpp
```diff
@@ -58,6 +59,78 @@ class BuffersTest : public ::testing::Test {
...
+        std::random_shuffle(mem.begin(), mem.end());
+
+        auto vec = std::unique_ptr<vec::Vector<T>>(new vec::Vector<T>(gf, n));
+        for (size_t i = 0; i < n; i++) {
+            vec->set(i, mem[i]);
+        }
+        return vec;
+    }
+
+    bool check_eq(T* buf1, T* buf2, size_t len)
```
slaperche-scality: You can probably use `memcmp` or `std::equal` for that.

slaperche-scality: `buf1` and `buf2` should be const

test/buffers_utest.cpp
```diff
@@ -58,6 +59,78 @@ class BuffersTest : public ::testing::Test {
...
+    {
+        for (size_t i = 0; i < len; ++i) {
+            if (buf1[i] != buf2[i]) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    bool check_all_zeros(T* buf, size_t len)
```
slaperche-scality: `buf` should be const

test/buffers_utest.cpp
```diff
@@ -58,6 +59,78 @@ class BuffersTest : public ::testing::Test {
...
+        for (size_t i = 0; i < len; ++i) {
+            if (buf[i] != 0) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    bool check_shuffled_bufs(
+        vec::Buffers<T>& input,
```
slaperche-scality: `input` and `output` should be const

src/fft_2n.h
```diff
@@ -140,6 +142,26 @@ void Radix2<T>::init_bitrev()
     }
 }
 
+template <typename T>
+void Radix2<T>::reverse(vec::Vector<T>& vec)
+{
+    for (unsigned i = 0; i < (unsigned)this->n; ++i) {
```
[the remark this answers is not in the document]
lamphamsy: `std::reverse` does different thing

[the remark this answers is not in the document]
slaperche-scality: Then your function should probably be renamed, otherwise it's confusing if your reverse is not exactly a reverse.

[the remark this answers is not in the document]
lamphamsy: okay, I will rename it

test/buffers_utest.cpp
```diff
@@ -80,7 +80,7 @@ class BuffersTest : public ::testing::Test {
 
     bool check_eq(const T* buf1, const T* buf2, size_t len)
     {
-        return memcmp(buf1, buf2, len * sizeof(T));
+        return (memcmp(buf1, buf2, len * sizeof(T)) == 0);
```
slaperche-scality: outer parenthesis are useless

You can write: `return memcmp(buf1, buf2, len * sizeof(T)) == 0`

src/fft_2n.h
slaperche-scality: Still here…

test/buffers_utest.cpp
```diff
@@ -58,6 +59,73 @@ class BuffersTest : public ::testing::Test {
...
+    {
+        std::vector<T> mem;
+        mem.reserve(max_val);
+
+        for (size_t i = 0; i < max_val; i++) {
+            mem.push_back(i);
+        }
+        std::random_shuffle(mem.begin(), mem.end());
+
+        auto vec = std::unique_ptr<vec::Vector<T>>(new vec::Vector<T>(gf, n));
```
slaperche-scality: std::make_unique

lamphamsy: done

PR #237, opened 2018-08-31
FFT::Radix2 uses bit-reversed permutation algorithm

lamphamsy:
Implement FFT Radix2 using bit-reversal permutation algorithm.

1. Decimation-in-time FFT
- Output is initialized as the input in the bit-reversal ordering
- Cooley-Tukey butterfly is performed on the output

It leads to an O(N*logN) complexity.

Note, we consider particular cases where the input is padded by zeros that always happens in the erasure encoding process. Normally, FFT process starts performing butterfly operations on groups of size 2, then 4, 8 etc. Thanks to zero padding, we can start from a group size where there is at most a non-zero element of input in the group, all other elements of the group are zero. Indeed, after performing butterfly operation on the group, all elements equal to the non-zero one.

Given FFT length N, a number of data length K, the output is initialized in the following clever way:
- Each group contains N/K elements
- Perform on groups containing a non-zero element of input: copy input's element at the bit-reversed index to all elements of the correspondent group of output
- For other groups, initialize them normally

It leads to an O(N*logK) complexity

2. Decimation-in-frequency FFT or inverse FFT
- Gentleman-Sande butterfly is performed on the input
- Output is copied from input at bit-reversed indices

It leads to an O(N*logN) complexity.

### DIFF
--- a/documentation/references.bib
+++ b/documentation/references.bib
@@ -41,3 +41,9 @@
     month     = {Jan},
     isbn      = {9782729852184}
 }
+
+@book{primenumbers,
+    author = {Richard Crandall and Carl Pomerance and Richard Crandall and Carl Pomerance},
+    title = {Prime numbers: a computational perspective. Second Edition},
+    year = {2005}
+}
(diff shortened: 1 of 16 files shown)


lamphamsy:
I reopen a PR for the branch.

lamphamsy:
Merging it

PR #238, opened 2018-08-31
Eh/clean property

lamphamsy:
A property carries extra-information (whose interpretation is left to the reader) related to a specific value (identified by its location).

It wraps a map whose each element is a key/value where
- key indicates the location of symbol whose value should be adjusted
- value is the value that could be used to adjust the symbol value

For prime fields, value is always 1.
For NF4, value is an uint32_t integer.


lamphamsy:
Reopen a PR for the branch

lamphamsy:
Merging it

Issue #222, opened 2018-07-27
Factorize SIMD code


slaperche-scality:
There is way too much code duplication/copy paste between all the SIMD implementation (`simd_{128,256}_u{16,32}.h`).

We **must** refactor this before adding more SIMD-feature, it's already barely maintainable…

lamphamsy:
It's fixed by #253 #254

PR #242, opened 2018-09-21
simplify the compile time option

slaperche-scality:
Instead of defining our own constants, we can use the "standard" ones
defined by the compiler.

Also, we don't need to infer the best SIMD instructions set on the
current machine by ourself: this can be done with `-march=native`.

Refs: #222


CMakeLists.txt
```diff
@@ -130,12 +124,15 @@ set(COMMON_CXX_FLAGS
 )
 
 # Option for enabling/disabling SIMD flags is for both of debug and release
-if (QUADIRON_USE_SIMD STREQUAL "SSE4")
+if (USE_SIMD STREQUAL "ON")
+  SET(COMMON_CXX_FLAGS "${COMMON_CXX_FLAGS} -march=native")
```
lamphamsy: Last time we do agree to not using `-march=native` ;) https://github.com/scality/quadiron/pull/190#discussion_r172463397

slaperche-scality: Yup, I said "IMHO, if we want to compile with march=native it should be done by enabling an option." and it's the case here (it's not always enabled, only when USE_SIMD is ON).

slaperche-scality:
> Do you test it on a machine that does not support AVX?

My laptop support AVX, but we do test the compilation without AVX (on Android). Why?

src/core.h
```diff
@@ -104,12 +104,10 @@ static inline std::mt19937& prng()
     return PRNG;
 }
 
-#ifdef QUADIRON_USE_SSE4
-#define ALIGN_SIZE 16
-#endif
-
-#ifdef QUADIRON_USE_AVX2
+#if defined(__AVX2__)
```
lamphamsy: If I understand correctly, this flag will be defined when the compiler option `-mavx2` is set. It does not depend on CPU's flag.
So I asked to run quadiron on a machine that does not support AVX to check that.

slaperche-scality: It's is defined if you ask for AVX **and** if you computer support it.

There is several cases to consider:
- you don't want any SIMD (e.g: to run a baseline benchmark): set `USE_SIMD` to `OFF`, `__AVX2__` won't be defined.
- you want the best SIMD available on your machine: set `USE_SIMD` to `ON`, `__AVX2__` will be defined only if your CPU supports it.
- you explicitely want AVX2 (but as you explicitely use it, you should know what you're doing): set `USE_SIMD` to `AVX` 

Basically, there is two high-level value for `USE_SIMD`: `ON` and `OFF`, and then there is the "expert" mode where you can select a specific instruction set yourself (but you should check that your CPU support it).

To be extra-safe, we could have a portable `cpuid` call that check which feature are available at runtime and abort with an error message if we see that the binary was compiled with an unsupported instruction set (that would be more user-friendly than dying with an Illegal Instruction), but that's just a nice to have.

lamphamsy: I cannot find a machine not support AVX
But by checking a similar flag for AVX512: compile with `-mavx512f` => `__AVX512F__` will be defined even my machine does not support it.

We support currently also the four cases you mentioned.

The checking of `cpuid` is made in `FindSIMD.cmake` that is deleted by this PR

slaperche-scality: > compile with -mavx512f => __AVX512F__ will be defined even my machine does not support it.

Yeah, I know that why I said "as you explicitely use it, you should know what you're doing".
Forcing the instruction set is not something you should do, if you do it you should know what you're doing.

> We support currently also the four cases you mentioned.

Yeah we support it but with a lot of custom code that duplicate what already exists in the compilers, and with some ugly logic where we parse by hand files like `cpuinfo`: it's messy and if we can avoid it, it's better.

> The checking of cpuid is made in FindSIMD.cmake that is deleted by this PR

That's not the kind of check I'm talking about.

What I removed here is a compile-time check, which can also generate binaries that abort with SIGILL so it's not that useful.
If we really want to prevent SIGILL we need a runtime check on the CPU, but IMHO this can be done later as a nice to have:

- the high level interface works fine (ON/OFF)
- the advanced interface (forcing an instruction set by hand) can generate binaries with SIGILL, but it's an advance interface (you shouldn't force an instruction set if you don't know what your CPU support: let the compiler do it for you).

lamphamsy: Alright, so we can go on this way :)

slaperche-scality:
@lamphamsy I kept `QUADIRON_USE_SIMD` because it's used to check if SIMD is enabled (that way we only have to check this one instead of checking if(SSE OR AVX OR AVX512 OR NEON), so it's convenient.

I only removed it's value (where 1 means SSE, 2 means AVX: now it's only defined without value and means "SIMD" is enabled => to see which SIMD, we can use the compiler's defines).

PR #244, opened 2018-10-01
fix TestFftAdd

slaperche-scality:
When `weak_rand` returns 2, the test fails with a FPE because we divide
by zero when calculating `t` in `test_taylor_expand`.

Note that fixing this is not enough, because the Taylor Expension expect
m >= 2 so anyway, we should execute theses test with `len = 2` (this is
properly handled, when calling `fft` but not when calling directly
`taylor_expand`.

Fix this by removing the call to `weak_rand` (it mades the test
non-reproducible anyway…).

### DIFF
--- a/test/fft_utest.cpp
+++ b/test/fft_utest.cpp
@@ -65,8 +65,9 @@ class FftTest : public ::testing::Test {
     void test_taylor_expand(const gf::Field<T>& gf, fft::Additive<T>* fft)
     {
         for (int i = 0; i < 1000; i++) {
-            const int t = 2 + gf.weak_rand() % (fft->get_n() - 2);
-            const int n = t + 1 + gf.weak_rand() % (fft->get_n() - t);
+            const int t = 2 + gf.weak_rand() % (fft->get_n() - 1);
+            const int n = t + gf.weak_rand() % (fft->get_n() - t + 1);
+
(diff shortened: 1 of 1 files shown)


test/fft_utest.cpp
```diff
@@ -295,7 +295,7 @@ TYPED_TEST(FftTest, TestFftAdd) // NOLINT
 
         unsigned len = this->code_len;
         if (gf.card_minus_one() <= this->code_len) {
-            len = gf.weak_rand();
+            len = gf_n;
```
lamphamsy: Okay, thanks :)

Please fix also L68: https://github.com/scality/quadiron/pull/244/files#diff-594f3543c3c5e448ed6d59ccc64d2169R68
```
            const int t = 2 + gf.weak_rand() % (fft->get_n() - 1);
            const int n = t + gf.weak_rand() % (fft->get_n() - t + 1);
```

slaperche-scality: This was to handle the case where `len = 2`, but if `len == 2` you'll fail while calling `taylor_expand` anyway, so this fix is not necessary, is it?

lamphamsy: It fixes another bug for `n` and `t`, as we want to have them such that `1 < t <= n`

slaperche-scality: Done

lamphamsy: thanks

Issue #225, opened 2018-08-03
Fix `hadamard_mul_doubled` in `simd_256_u32.h`


slaperche-scality:
Bug spotted by clang-analyzer:

```
quadiron/src/simd_256_u32.h:383:13: warning: Value stored to '_x_p' is never read [clang-analyzer-deadcode.DeadStores]
            _x_p = add(_x_p, _y_p, card);
            ^
quadiron/src/simd_256_u32.h:383:13: note: Value stored to '_x_p' is never read
quadiron/src/simd_256_u32.h:384:13: warning: Value stored to '_x_next_p' is never read [clang-analyzer-deadcode.DeadStores]
            _x_next_p = add(_x_next_p, _y_p, card);
            ^
src/simd_256_u32.h:384:13: note: Value stored to '_x_next_p' is never read
src/simd_256_u32.h:410:13: warning: Value stored to '_x_p' is never read [clang-analyzer-deadcode.DeadStores]
            _x_p = mul(_x_p, _y_p, F4);
            ^
quadiron/src/simd_256_u32.h:410:13: note: Value stored to '_x_p' is never read
quadiron/src/simd_256_u32.h:446:13: warning: Value stored to '_x_p' is never read [clang-analyzer-deadcode.DeadStores]
            _x_p = mul(_x_p, _y_p, F4);
            ^
quadiron/src/simd_256_u32.h:446:13: note: Value stored to '_x_p' is never read
quadiron/src/simd_256_u32.h:447:13: warning: Value stored to '_x_next_p' is never read [clang-analyzer-deadcode.DeadStores]
            _x_next_p = mul(_x_next_p, _y_p, F4);
            ^
/home/sylvain/dev/quadiron/src/simd_256_u32.h:447:13: note: Value stored to '_x_next_p' is never read
```

PR #245, opened 2018-10-04
Update README

slaperche-scality:
Since 5d19e845eb8b5c7e9df1cae9d6180f9bf7840613 we are now using C++14


slaperche-scality:
Lint fail is related to issue #225, not related to these patch

PR #250, opened 2018-10-11
add file checks and zpad options

vrancurel:
@slaperche-scality 2 problems here:
- I don't have the proper indentation, should we share the indent settings for emacs/vim that everybody use ?
- file exceptions don't seem to work

### DIFF
--- a/scripts/test_ec.sh
+++ b/scripts/test_ec.sh
@@ -68,19 +68,30 @@ do_test()
 
     rm -f foo.*
 
-    for i in `seq 0 $(expr ${n_data} - 1)`
+    for i in `seq -w 0 $(expr ${n_data} - 1)`
     do
         dd if=/dev/urandom of=foo.d${i} bs=${bs} count=1 > /dev/null 2>&1
         md5sum foo.d${i} > foo.d${i}.md5sum.1
+        checkfail "checksumming data"
     done
 
     echo -n "GEN,"
     ${valgrind} ${bin} -e ${fec_type} -w ${word_size} -n ${n_data} -m ${n_coding} -p foo -c ${extraopts} ${vflag}
     checkfail "coding generation"
 
-    for i in `seq 0 $(expr ${n_coding} - 1)`
+    if [ "${type}" = "NON_SYSTEMATIC" ]
+    then
+        _n_coding=$(expr ${n_data} + ${n_coding})
+        _n_digits=${#_n_coding}
+        _fmt=%0${_n_digits}d
+    else
+        _n_coding=${n_coding}
+    fi
+    
+    for i in `seq -w 0 $(expr ${_n_coding} - 1)`
     do
         md5sum foo.c${i} > foo.c${i}.md5sum.1
+        checkfail "checksumming codings"
     done
 
     if [ "${type}" = "NON_SYSTEMATIC" ]
@@ -92,35 +103,44 @@ do_test()
         for i in $files
         do
             mv ${i} ${i}.1
+            checkfail "mv data"
         done
     fi
 
-    j=0
-    for i in $data_loss
-    do
-        if [ "${type}" = "NON_SYSTEMATIC" ]
-        then
-            mv foo.c${j} foo.c${j}.1
-            mv foo.c${j}.props foo.c${j}.props.1
-        else
-            mv foo.d${i} foo.d${i}.1
-        fi
-        j=`expr ${j} + 1`
-    done
-
-    for i in $coding_loss
-    do
-        if [ "${type}" = "NON_SYSTEMATIC" ]
-        then
-            mv foo.c${j} foo.c${j}.1
-            mv foo.c${j}.props foo.c${j}.props.1
-        else
-            mv foo.c${i} foo.c${i}.1
-            mv foo.c${i}.props foo.c${i}.props.1
-        fi
-        j=`expr ${j} + 1`
-    done
-
+    if [ "${type}" = "NON_SYSTEMATIC" ]
+    then
+        for pos in $data_loss
+        do
+            _pos=$(printf ${_fmt} ${pos})
+            mv foo.c${_pos} foo.c${_pos}.1
+            checkfail "data loss: mv non-sys codings"
+            mv foo.c${_pos}.props foo.c${_pos}.props.1
+            checkfail "data loss: mv non-sys props"
+        done
+        for pos in $coding_loss
+        do
+            _pos=$(expr ${n_data} + ${pos})
+            _pos=$(printf ${_fmt} ${_pos})
+            mv foo.c${_pos} foo.c${_pos}.1
+            checkfail "coding loss: mv non-sys codings"
+            mv foo.c${_pos}.props foo.c${_pos}.props.1
+            checkfail "coding loss: nv non-sys props"
+        done
+    else
+        for pos in $data_loss
+        do
+            mv foo.d${pos} foo.d${pos}.1
+            checkfail "data loss: mv data"
+        done
+        for pos in $coding_loss
+        do
+            mv foo.c${pos} foo.c${pos}.1
+            checkfail "coding loss: nm codings"
+            mv foo.c${pos}.props foo.c${pos}.props.1
+            checkfail "coding loss: mv props"
+        done
+    fi
+    
     if [ "${directive}" == "enconly" ]
     then
         echo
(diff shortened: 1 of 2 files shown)


slaperche-scality:
> I don't have the proper indentation, should we share the indent settings for emacs/vim that everybody use ?

The settings are defined by `.clang-format`, which is used by `clang-format`. You can manually run it using `make format`.
You can also make this more automatic by:
- configuring your editor to run the command each time a file is saved
- using a git hook to run a command each time you commit (pre-commit hook).

> file exceptions don't seem to work

By default, C++ doesn't raise exceptions when error ocurs on I/O operations: instead, it sets [flags](http://www.cplusplus.com/reference/ios/ios/rdstate/) on the stream that you need to check manually

```C++
#include <iostream>
#include <fstream>

int main () {
    std::ifstream fp("test.txt");
    if ((fp.rdstate() & std::ifstream::failbit) != 0) {
        std::cerr << "Error while opening the file\n";
    }
    return 0;
}
```

The test can be simplified by using the helper `fail`: `if (fp.fail())`.
Or even more, by taking advantage of the fact that the stream implement a convertion to `bool` that returns false if the stream is in an error state (see [here](http://www.cplusplus.com/reference/ios/ios/operator_bool/)) so you can even write:

```C++
#include <iostream>   
#include <fstream>

int main () {
    std::ifstream fp("test.txt");
    if (!fp) {
        std::cerr << "Error while opening the file\n";
    }
    return 0;
}
```

Which is like what you do in C when testing if `fopen` doesn't return NULL.

Now, if you want the errors to raise instead of setting a flag, it's opt-in and can be configured using [exceptions](https://en.cppreference.com/w/cpp/io/basic_ios/exceptions) to set the exception mask.

Example:

```C++
#include <iostream>   
#include <fstream>

int main () {
    std::ifstream fp;

    fp.exceptions(std::ifstream::failbit);
    try {
        fp.open("test.txt");
    } catch (const std::ifstream::failure& exn) {
        std::cerr << "Error while opening the file\n";
    }
    return 0;
}

```

test/ec_driver.cpp
```diff
@@ -78,23 +79,38 @@ void create_coding_files(
...
-        snprintf(filename, sizeof(filename), "%s.d%d", prefix, i);
+        snprintf(filename, sizeof(filename), "%s.d%0*d", prefix, zpad, i);
         if (vflag)
             std::cerr << "create: opening data " << filename << "\n";
-        d_files[i] = new std::ifstream(filename);
+        try {
+          d_files[i] = new std::ifstream(filename);
+        } catch (const std::ifstream::failure& e) {
+          std::cerr << "Exception opening " << filename << "\n";
+          exit(1);
```
slaperche-scality: `std::exit(EXIT_FAILURE);` would be cleaner/more explicit then `exit(1)` (may need an `#include <cstdlib>`).

Same remarks for the other occurrences.

test/ec_driver.cpp
```diff
@@ -78,23 +79,38 @@ void create_coding_files(
     std::vector<quadiron::Properties> c_props(fec->n_outputs);
 
     for (unsigned i = 0; i < fec->n_data; i++) {
-        snprintf(filename, sizeof(filename), "%s.d%d", prefix, i);
+        snprintf(filename, sizeof(filename), "%s.d%0*d", prefix, zpad, i);
```
slaperche-scality: Not a big fan of `snprintf` here, it's more C than C++ and it can truncate the output (and we don't check the return value).

This formatting could be done using a `stringstream`. Example:

```C++
#include <cstdio>
#include <iomanip>
#include <iostream>
#include <sstream>


static inline
std::string get_filename(const std::string& pfx, int part, int zpad)
{
    std::ostringstream oss;
    oss << pfx << ".d" << std::setfill('0') << std::setw(zpad) << part;
    return oss.str();
}

int main () {
    int zpad = 10;
    int part = 42;
    char filename[1024];
    const char* prefix = "foo";

    snprintf(filename, sizeof(filename), "%s.d%0*d", prefix, zpad, part);
    std::cout << filename << '\n';

    std::cout << get_filename(prefix, part, zpad) << '\n';
    return 0;
}
```

Same for the other occurrences.

test/ec_driver.cpp
```diff
@@ -137,39 +153,59 @@ bool repair_data_files(
...
         if (vflag)
             std::cerr << "repair: checking coding " << filename << "\n";
         if (access(filename, F_OK)) {
             if (vflag)
                 std::cerr << filename << " is missing\n";
             c_files[i] = nullptr;
         } else {
+          try {
             c_files[i] = new std::ifstream(filename);
+          } catch (const std::ifstream::failure& e) {
```
slaperche-scality: We really have a lot of duplicated code/copy-paste in this file (all the loop to open streams), this could probably be factored in an helper function.

A simple change like that shouldn't require so many modifications :/

test/ec_driver.cpp
vrancurel: OK thanks for the comments, I will address this.

vrancurel:
I did not implement the snprintf replacement because I was not sure your get_filename() does not return something not allocated (allocated in the stack only).

Now the zpad is computed automatically which makes it compatible with unix tools like split

E.g. splitting a video file:

`
split file1.mp4 -b1M -d x.d
truncate --size 1048576 x.d89  #for now the latest video chunk has to be truncated
ec_driver -e rs-fnt -w 2 -n 90 -m 160 -p x -c 
rm x.d*
rm x.c1?? x.c1??.props
ec_driver -e rs-fnt -w 2 -n 90 -m 160 -p x -r 
cat x.d* > file2.mp4
`

slaperche-scality:
> I did not implement the snprintf replacement because I was not sure your get_filename() does not return something not allocated (allocated in the stack only).

It's safe.
The string object itself is indeed allocated on the stack and returned by value, but its content lives on the heap (except when SSO optimization kicks in, but even so it's not an issue because then it's copied along the string object when returned).

> Now the zpad is computed automatically which makes it compatible with unix tools like split
Ho, that's nice :)

test/ec_driver.cpp
```diff
@@ -62,6 +64,14 @@ char* xstrdup(const char* str)
     return n;
 }
 
+u_int ilog10(u_int v)
+{
+  return (v >= 1000000000u) ? 9 : (v >= 100000000u) ? 8 :
```
slaperche-scality: nested ternary are not really readable nor easy to debug.

I would either use explicit `if` or, since this is not a critical codepath I would use the naive approach:

```c++
unsigned count_digits(uint32_t number)
{
    int digits = 0;
    while (number) {
        number /= 10;
        digits++;
    }
    return digits;
}
```

Bonus point: now the caller don't need to add +1 itself, it's directly handled here.

test/ec_driver.cpp
```diff
@@ -78,23 +88,35 @@ void create_coding_files(
     std::vector<quadiron::Properties> c_props(fec->n_outputs);
 
     for (unsigned i = 0; i < fec->n_data; i++) {
-        snprintf(filename, sizeof(filename), "%s.d%d", prefix, i);
+        snprintf(filename, sizeof(filename), "%s.d%0*d", prefix, data_zpad, i);
```
slaperche-scality: Can be replaced, the alternative I proposed is safe.

vrancurel:
@slaperche-scality can you review plz ?

test/ec_driver.cpp
```diff
@@ -71,30 +97,42 @@ void create_coding_files(
...
     std::vector<quadiron::Properties> c_props(fec->n_outputs);
 
     for (unsigned i = 0; i < fec->n_data; i++) {
-        snprintf(filename, sizeof(filename), "%s.d%d", prefix, i);
+        filename = get_filename(prefix, 'd', data_zpad, i, "");
         if (vflag)
             std::cerr << "create: opening data " << filename << "\n";
         d_files[i] = new std::ifstream(filename);
+        if (d_files[i]->fail()) {
+            std::cerr << "Exception opening " << filename << "\n";
```
slaperche-scality: It would be nice to have different error message for the different case (so that looking at the error message, you know where piece of code failed).
e.g:

- l.112: "Exception opening data file (RO) for creation "
- l.124: "Exception opening coding file for creation "
- l.133: "Exception opening coding props for creation "
- l.187: "Exception opening data file for checking "
- l.194: "Exception opening data file (RO) for checking "
- l.211: "Exception opening coding file for checking "
- l.223: "Exception opening coding props for checking "

scripts/test_ec.sh
```diff
@@ -68,19 +68,30 @@ do_test()
 
     rm -f foo.*
 
-    for i in `seq 0 $(expr ${n_data} - 1)`
+    for i in `seq -w 0 $(expr ${n_data} - 1)`
```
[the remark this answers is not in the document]
vrancurel: The zero-padding affects the test, so we have to zero pad also in the test.

vrancurel:
@slaperche-scality shall be good

PR #252, opened 2018-10-15
Slight enhancement

lamphamsy:
- Systematic RS-FNT: decode intermediate symbols using Radix-2 FFT on a shorten-length input
- Pack/unpack: don't need to create temporary vectors
- Enhance slightly the multiplication of a vector to a Buffers to avoid as much as possible multiplying two numbers
- FEC utest: add test for FNT with word size = 1 and 2

### DIFF
--- a/src/vec_cast.h
+++ b/src/vec_cast.h
@@ -51,13 +51,10 @@ inline void pack_next(
     int n,
     size_t size)
 {
-    int i;
-    std::vector<Tw*> tmp(n, nullptr);
-    for (i = 0; i < n; i++) {
-        tmp[i] = reinterpret_cast<Tw*>(src.at(i));
-        std::copy_n(tmp[i], size, dest.at(i));
+    for (int i = 0; i < n; i++) {
+        Tw* tmp = reinterpret_cast<Tw*>(src.at(i));
+        std::copy_n(tmp, size, dest.at(i));
     }
-    tmp.shrink_to_fit();
 }
 
 /*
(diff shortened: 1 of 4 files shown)


src/vec_cast.h
```diff
@@ -51,13 +51,10 @@ inline void pack_next(
     int n,
     size_t size)
 {
-    int i;
-    std::vector<Tw*> tmp(n, nullptr);
-    for (i = 0; i < n; i++) {
-        tmp[i] = reinterpret_cast<Tw*>(src.at(i));
-        std::copy_n(tmp[i], size, dest.at(i));
+    for (int i = 0; i < n; i++) {
+        Tw* tmp = reinterpret_cast<Tw*>(src.at(i));
```
slaperche-scality: What can be the types of `Tw` and `Td` here?

(Just to be sure that we aren't breaking the strict aliasing rule)

src/vec_cast.h
```diff
@@ -85,16 +82,24 @@ inline void pack(
     assert(sizeof(Td) >= word_size);
     assert(word_size % sizeof(Ts) == 0);
     // get only word_size bytes from each element
-    if (word_size == 1) {
+    switch (word_size) {
```
slaperche-scality: Out of the scope of this PR, but I wonder if `word_size` could be a template parameter.

In other word, does it make sense to change word_size after instanciating the FEC object?

src/vec_cast.h
slaperche-scality: Then I think that we have an issue with strict aliasing violation here.

Not introduced by this PR so we can fix it later, but maybe open an issue about it so we don't forget.

PR #253, opened 2018-10-15
Part1: Radix-2 FFT enhancement by working on two layers per iteration

lamphamsy:
The base branch `eh/fft_radix2_and_re_work_sim` would be merged into master when two parts are done: #253 #254 

## Part1

- Butterly CT on two-layers at a time

For each quadruple `(P, Q, R, S) = (buf[i], buf[i + m], buf[i + 2 * m], buf[i + 3 * m])`

First layer: butterfly on `(P, Q)` and` (R, S)` for `step = 2 * m`

```
      coef r1 = W[start * n / (2 * m)]
      P = P + r1 * Q
      Q = P - r1 * Q
      R = R + r1 * S
      S = R - r1 * S
```

Second layer: butterfly on `(P, R)` and `(Q, S)` for `step = 4 * m`

```
      coef r2 = W[start * n / (4 * m)]
      coef r3 = W[(start + m) * n / (4 * m)]
      P = P + r2 * R
      R = P - r2 * R
      Q = Q + r3 * S
      S = Q - r3 * S
```

- Remove useless butterfly operations on GF


slaperche-scality:
Your explanations are nice, IMHO you should also put them in the message of the relevant commit.

That way they can be accessible directly from Git log (no dependence to github, can dive into the code offline, …).

src/fft_2n.h
```diff
@@ -336,39 +488,135 @@ void Radix2<T>::fft(vec::Buffers<T>& output, vec::Buffers<T>& input)
...
+        T* a = buf.get(i);
+        T* b = buf.get(i + m);
+        // perform butterfly operation for Cooley-Tukey FFT algorithm
+        for (size_t j = 0; j < this->pkt_size; ++j) {
+            b[j] = this->gf->mul(coef, a[j]);
+        }
+    }
+}
+
+// for each pair (P, Q) = (buf[i], buf[i + m]):
```
slaperche-scality: Can't we merge `butterfly_gs_step` with `butterfly_gs_step_offset` by using a default value for `offset`, i.e

```
template <typename T>
void Radix2<T>::butterfly_gs_step_offset(
    vec::Buffers<T>& buf,
    T coef,
    unsigned start,
    unsigned m,
    unsigned step,
    size_t offset = 0)
```

BTW,  `butterfly_gs_step_offset` seems unused, so it can be removed, no?

src/fft_2n.h
```diff
@@ -284,39 +339,136 @@ template <typename T>
...
+}
 
-                // perform butterfly operation for Cooley-Tukey FFT algorithm
-                this->gf->butterfly_ct(r, a, b, size);
-            }
+// for each pair (P, Q) = (buf[i], buf[i + m]):
+// P = P + c * Q
+// Q = P - c * Q
+template <typename T>
+void Radix2<T>::butterfly_ct_step(
```
slaperche-scality: The code seems identical to `butterfly_ct_step_offset`, with `offset=0`, so I think the code could be merged by giving offset a default value of `0`.

Or imply get rid of `butterfly_ct_step_offset`, as it seems unused.

src/fft_2n.h
lamphamsy: `butterfly_gs_step_offset` is used in specialized functions: `fft_2n.cpp`.
But I will merge them to avoid duplicated codes.

lamphamsy: - [ ] Remove duplicated codes

src/fft_2n.h
lamphamsy: I aggregated `butterfly_ct_step` and `butterfly_ct_step_offset` :)

lamphamsy:
@slaperche-scality : Comments are addressed. Could you review the PR? Thanks

src/fft_2n.h
slaperche-scality: Why having

```
template <typename T>
void Radix2<T>::butterfly_gs_step_simple(
    vec::Buffers<T>& buf, T coef, unsigned start, unsigned m, unsigned step
)
{
    butterfly_gs_step_offset_simple(buf, coef, start, m, step, 0);
}

template <typename T>
void Radix2<T>::butterfly_gs_step_offset_simple(
    vec::Buffers<T>& buf, T coef, unsigned start, unsigned m, unsigned step, size_t offset
)
{
    […]
}
```

instead of

```
template <typename T>
void Radix2<T>::butterfly_gs_step_simple(
    vec::Buffers<T>& buf, T coef, unsigned start, unsigned m, unsigned step, size_t offset=0
)
{
    […]
}
```

?

Using default arguments, you don't need two functions: one is enough.

lamphamsy: The reason is the same for `butterfly_gs_step_offset` as I said above: `butterfly_gs_step_simple` will be specialized, and it will use `butterfly_gs_step_offset_simple` to avoid duplication.
These specialized functions are in `fft_2n.cpp`, you can see it in #254

slaperche-scality: I don't understand, you still can have a single function if you use a default argument.

Did you tried?

I tried (with the following diff, if it can help you) and it works!

```diff
diff --git a/src/fft_2n.cpp b/src/fft_2n.cpp
index dea8e46..efa43ca 100644
--- a/src/fft_2n.cpp
+++ b/src/fft_2n.cpp
@@ -70,18 +70,18 @@ void Radix2<uint16_t>::butterfly_ct_two_layers_step(
         //  ---------
         const uint16_t r1 = W->get(start * this->n / m / 2);
         // first pair
-        butterfly_ct_step_offset(buf, r1, start, m, step, offset);
+        butterfly_ct_step(buf, r1, start, m, step, offset);
         // second pair
-        butterfly_ct_step_offset(buf, r1, start + 2 * m, m, step, offset);
+        butterfly_ct_step(buf, r1, start + 2 * m, m, step, offset);
         //  ---------
         // Second layer
         //  ---------
         // first pair
         const uint16_t r2 = W->get(start * this->n / m / 4);
-        butterfly_ct_step_offset(buf, r2, start, 2 * m, step, offset);
+        butterfly_ct_step(buf, r2, start, 2 * m, step, offset);
         // second pair
         const uint16_t r3 = W->get((start + m) * this->n / m / 4);
-        butterfly_ct_step_offset(buf, r3, start + m, 2 * m, step, offset);
+        butterfly_ct_step(buf, r3, start + m, 2 * m, step, offset);
     }
 }
 
@@ -104,7 +104,7 @@ void Radix2<uint16_t>::butterfly_ct_step(
     // for last elements, perform as non-SIMD method
     if (last_len > 0) {
         size_t offset = vec_len * ratio;
-        butterfly_ct_step_offset(buf, r, start, m, step, offset);
+        butterfly_ct_step(buf, r, start, m, step, offset);
     }
 }
 
@@ -182,18 +182,18 @@ void Radix2<uint32_t>::butterfly_ct_two_layers_step(
         //  ---------
         const uint32_t r1 = W->get(start * this->n / m / 2);
         // first pair
-        butterfly_ct_step_offset(buf, r1, start, m, step, offset);
+        butterfly_ct_step(buf, r1, start, m, step, offset);
         // second pair
-        butterfly_ct_step_offset(buf, r1, start + 2 * m, m, step, offset);
+        butterfly_ct_step(buf, r1, start + 2 * m, m, step, offset);
         //  ---------
         // Second layer
         //  ---------
         // first pair
         const uint32_t r2 = W->get(start * this->n / m / 4);
-        butterfly_ct_step_offset(buf, r2, start, 2 * m, step, offset);
+        butterfly_ct_step(buf, r2, start, 2 * m, step, offset);
         // second pair
         const uint32_t r3 = W->get((start + m) * this->n / m / 4);
-        butterfly_ct_step_offset(buf, r3, start + m, 2 * m, step, offset);
+        butterfly_ct_step(buf, r3, start + m, 2 * m, step, offset);
     }
 }
 
@@ -216,7 +216,7 @@ void Radix2<uint32_t>::butterfly_ct_step(
     // for last elements, perform as non-SIMD method
     if (last_len > 0) {
         size_t offset = vec_len * ratio;
-        butterfly_ct_step_offset(buf, r, start, m, step, offset);
+        butterfly_ct_step(buf, r, start, m, step, offset);
     }
 }
 
diff --git a/src/fft_2n.h b/src/fft_2n.h
index 68f83b1..8829d9e 100644
--- a/src/fft_2n.h
+++ b/src/fft_2n.h
@@ -94,7 +94,8 @@ class Radix2 : public FourierTransform<T> {
         T r,
         unsigned start,
         unsigned m,
-        unsigned step);
+        unsigned step,
+        size_t offset=0);
     void butterfly_ct_two_layers_step(
         vec::Buffers<T>& buf,
         unsigned start,
@@ -113,13 +114,6 @@ class Radix2 : public FourierTransform<T> {
         unsigned step);
 
     // Only used for non-vectorized elements
-    void butterfly_ct_step_offset(
-        vec::Buffers<T>& buf,
-        T coef,
-        unsigned start,
-        unsigned m,
-        unsigned step,
-        size_t offset);
     void butterfly_gs_step_offset(
         vec::Buffers<T>& buf,
         T coef,
@@ -380,29 +374,6 @@ void Radix2<T>::fft(vec::Buffers<T>& output, vec::Buffers<T>& input)
     }
 }
 
-// for each pair (P, Q) = (buf[i], buf[i + m]):
-// P = P + c * Q
-// Q = P - c * Q
-template <typename T>
-void Radix2<T>::butterfly_ct_step(
-    vec::Buffers<T>& buf,
-    T r,
-    unsigned start,
-    unsigned m,
-    unsigned step)
-{
-    for (int i = start; i < this->n; i += step) {
-        T* a = buf.get(i);
-        T* b = buf.get(i + m);
-        // perform butterfly operation for Cooley-Tukey FFT algorithm
-        for (size_t j = 0; j < this->pkt_size; ++j) {
-            T x = this->gf->mul(r, b[j]);
-            b[j] = this->gf->sub(a[j], x);
-            a[j] = this->gf->add(a[j], x);
-        }
-    }
-}
-
 /**
  * Butterly CT on two-layers at a time
  *
@@ -453,7 +424,7 @@ void Radix2<T>::butterfly_ct_two_layers_step(
 }
 
 template <typename T>
-void Radix2<T>::butterfly_ct_step_offset(
+void Radix2<T>::butterfly_ct_step(
     vec::Buffers<T>& buf,
     T coef,
     unsigned start,
@@ -642,7 +613,8 @@ void Radix2<uint16_t>::butterfly_ct_step(
     uint16_t r,
     unsigned start,
     unsigned m,
-    unsigned step);
+    unsigned step,
+    size_t offset=0);
 template <>
 void Radix2<uint16_t>::butterfly_gs_step(
     vec::Buffers<uint16_t>& buf,
@@ -669,7 +641,8 @@ void Radix2<uint32_t>::butterfly_ct_step(
     uint32_t r,
     unsigned start,
     unsigned m,
-    unsigned step);
+    unsigned step,
+    size_t offset=0);
 template <>
 void Radix2<uint32_t>::butterfly_gs_step(
     vec::Buffers<uint32_t>& buf,
```

lamphamsy: Thank you. I have not tried it :)
In your diff of fft_2n.cpp, I observe infinite loops, e.g.
```
@@ -104,7 +104,7 @@ void Radix2<uint16_t>::butterfly_ct_step(
     // for last elements, perform as non-SIMD method
     if (last_len > 0) {
         size_t offset = vec_len * ratio;
-        butterfly_ct_step_offset(buf, r, start, m, step, offset);
+        butterfly_ct_step(buf, r, start, m, step, offset);
     }
 }
```
I'm wondering if it works. I will check it.

lamphamsy: Nope, it does not work :(

slaperche-scality: Really? (beware, I made this diff on top of you PR, not on this commit specifically).
Because I ran `make check` and all the test were passing.

lamphamsy: it's possible to handle the trailing bits in SIMD part, but these functions are very similar to template member functions. Anyway, I'm waiting for your comment :)

slaperche-scality: Yes, you're right :)

I thought about it and it will only move the problem, not solve it.

I can propose you the following:

We define the default slow operation in a well-named method `butterfly_gs_step_simple_slow`

```c++
template <typename T>
void Radix2<T>::butterfly_gs_step_simple_slow(
    vec::Buffers<T>& buf,
    T coef,
    unsigned start,
    unsigned m,
    unsigned step,
    size_t offset=0)
{
    for (int i = start; i < this->n; i += step) {
        T* a = buf.get(i);
        T* b = buf.get(i + m);
        // perform butterfly operation for Cooley-Tukey FFT algorithm
        for (size_t j = offset; j < this->pkt_size; ++j) {
            T x = this->gf->sub(a[j], b[j]);
            a[j] = this->gf->add(a[j], b[j]);
            b[j] = this->gf->mul(coef, x);
        }
    }
}
```

The we have the default template definition that relies on this slow implementation:

```c++
template <typename T>
void Radix2<T>::butterfly_gs_step_simple(
    vec::Buffers<T>& buf, T coef, unsigned start, unsigned m, unsigned step
)
{
    butterfly_gs_step_simple_slow(buf, coef, start, m, step);
}
```

And finally we have the specialized SIMD version that use faster code for the main processing and fallback on the slow implementation for the trailing elements.

```c++
template <>
void Radix2<uint16_t>::butterfly_gs_step_simple(
    vec::Buffers<uint16_t>& buf,
    uint16_t coef,
    unsigned start,
    unsigned m,
    unsigned step)
{
    const unsigned ratio = ALIGN_SIZE / sizeof(uint16_t);
    const size_t len = this->pkt_size;
    const size_t vec_len = len / ratio;
    const size_t last_len = len - vec_len * ratio;

    // perform vector operations
    simd::butterfly_gs_step_simple(buf, coef, start, m, vec_len, card);

    // for last elements, perform as non-SIMD method
    if (last_len > 0) {
        size_t offset = vec_len * ratio;
        butterfly_gs_step_simple_slow(buf, coef, start, m, step, offset);
    }
}
```

What do you think?

lamphamsy: It's exactly what I did, except the function name and the optional parameter `offset` :)

slaperche-scality: Yup, pretty similar to what you did, but the naming `slow` is more explicit and naming is important :)
(and the default parameter avoid to specify the offset when you don't need it => but that's minor)

lamphamsy: ok, I will update it :)

PR #254, opened 2018-10-15
Part2: Vectorize operations for Radix-2 FFT & re-work vectorization part

lamphamsy:
The target branch `eh/fft_radix2_and_re_work_sim` would be merged into master when two parts are done: #253 #254 

## Part2

- Vectorize operations for Radix-2 FFT
- Re-work also vectorization part to avoid duplication codes


src/simd_fnt.h
```diff
@@ -0,0 +1,555 @@
...
+    unsigned start,
+    unsigned m,
+    unsigned step,
+    size_t len,
+    T card)
+{
+    if (len == 0) {
+        return;
+    }
+    const T rp1 = r + 1;
```
slaperche-scality: what does `rp` mean?

src/simd_fnt.h
```diff
@@ -0,0 +1,555 @@
...
+        return x;
+    } else if (rp1 < q) {
+        return MUL_MOD(c, x, q);
+    } else {
+        return NEG_MOD(x, q);
+    }
+}
+
+/**
+ * Vectorized butterly CT step
```
slaperche-scality: butter**F**ly (typo is present in other places in the file).

src/simd_fnt.h
```diff
@@ -0,0 +1,555 @@
...
+{
+    if (len == 0) {
+        return;
+    }
+    const T rp1 = r + 1;
+    VecType c = SET1(r);
+
+    const size_t end = (len > 1) ? len - 1 : 0;
+    const unsigned bufs_nb = buf.get_n();
+    // #pragma omp parallel for
```
slaperche-scality: Should be removed:
- we are not using OpenPM yet
- this is dead code (it's commented)
- it's not related to this PR

Same comment for all the other occurrences.

src/simd_fnt.h
```diff
@@ -0,0 +1,555 @@
...
+    // #pragma unroll
+    const std::vector<T*>& mem = buf.get_mem();
+    for (unsigned i = start; i < bufs_nb; i += step) {
+        VecType x1, y1;
+        VecType x2, y2;
+        VecType* __restrict p = reinterpret_cast<VecType*>(mem[i]);
```
slaperche-scality: I think we should avoid restrict (see my comment about it), or if we really need it then be extra-careful.

src/fft_2n.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+    // for last elements, perform as non-SIMD method
+    if (last_len > 0) {
+        size_t offset = vec_len * ratio;
+        butterfly_gs_step_offset(buf, coef, start, m, step, offset);
+    }
+}
+
+template <>
+void Radix2<uint16_t>::butterfly_gs_step_simple(
```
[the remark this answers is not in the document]
lamphamsy: No, the functions do different things. Why do you say that they contain exactly the same code?
For the trailing elements, as we discussed in #253 , we keep it here.

src/simd_128.h
```diff
@@ -33,19 +33,131 @@
...
-const m128i F4_m128i = _mm_set1_epi32(65537);       // NOLINT(cert-err58-cpp)
-const m128i F4minus1_m128i = _mm_set1_epi32(65536); // NOLINT(cert-err58-cpp)
-const m128i F3_m128i = _mm_set1_epi32(257);         // NOLINT(cert-err58-cpp)
-const m128i F3minus1_m128i = _mm_set1_epi32(256);   // NOLINT(cert-err58-cpp)
+typedef __m128i VecType;
+typedef uint32_t MaskIntType;
 
-const m128i F3_m128i_u16 = _mm_set1_epi16(257);       // NOLINT(cert-err58-cpp)
-const m128i F3minus1_m128i_u16 = _mm_set1_epi16(256); // NOLINT(cert-err58-cpp)
+#define F4_u32 _mm_set1_epi32(65537)
```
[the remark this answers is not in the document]
lamphamsy: It's to avoid the NOLINT comment only :)

src/simd_128.h
```diff
@@ -33,19 +33,131 @@
...
+
+inline VecType CMPEQ32(VecType x, VecType y)
+{
+    return _mm_cmpeq_epi32(x, y);
+}
+inline VecType CMPGT32(VecType x, VecType y)
+{
+    return _mm_cmpgt_epi32(x, y);
+}
+inline VecType MINU32(VecType x, VecType y)
```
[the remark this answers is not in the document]
lamphamsy: `u` stands for `unsigned`

src/simd_128.h
```diff
@@ -33,19 +33,131 @@
...
+
+inline VecType AND(VecType x, VecType y)
+{
+    return _mm_and_si128(x, y);
+}
+inline VecType XOR(VecType x, VecType y)
+{
+    return _mm_xor_si128(x, y);
+}
+inline VecType SHIFTR_1(VecType x)
```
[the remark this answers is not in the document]
lamphamsy: I tried this approach, but there will raise an error in compiling as the 2nd argument of `_mm_srli_si128` must be explicitly a constant, e.g. 1, 2 etc.

src/fft_2n.cpp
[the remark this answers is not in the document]
slaperche-scality: Yup, for trailing element we can forget it (I started the review this morning and didn't went back to edit the comment).

What are the different thing they are doing? They iterate over `buf` and call a function `f` and only `f` changes, no?

src/simd_basic.h
```diff
@@ -0,0 +1,404 @@
...
+ * Modular general multiplication for packed unsigned 32-bit integers
+ *
+ * @note It's necessary to verify overflow on multiplying elements
+ *
+ * @param x input register
+ * @param y input register
+ * @param q modulo
+ * @return (x * y) mod q
+ */
+inline VecType MULFULL_MOD(VecType x, VecType y, uint32_t q)
```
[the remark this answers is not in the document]
lamphamsy: This is micro-optimization as it's crucial for performance. `MUL_MOD` is used in FNT, while `MULFULL_MOD` in NF4

src/simd_128.h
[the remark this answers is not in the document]
slaperche-scality: Not a good reason.

const are typed, preprocessor constant aren't.
const can be printed in GDB, preprocessor constant no (except with some debugging option).
…

So, I would rather keep the NOLINT and the const.

Moreover, the NOLINT can be removed for recent Clang (>= 6), so no reason to replace the const by a define.

src/simd_128.h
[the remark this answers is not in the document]
slaperche-scality: But we aren't using `_mm_min_epi32`, so there is no need to make the name less clear to have a distinction with something unused.

src/simd_basic.h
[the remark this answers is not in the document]
lamphamsy: I test for FNT with your proposition, i.e. using optional argument `filter`: encoding speed ~2050MB/s
My current implementation without `filter` gives about ~2150MB/s
I have tested the template parameter way yet.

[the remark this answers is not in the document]
slaperche-scality: So almost 5%, quite costly…
A drop in perf was to be expected because of the branch that may be hard to predict, the template version should perform better.

lamphamsy:
@slaperche-scality could you review it?

slaperche-scality:
Could you squash all the "rename *" commits into the one that introduced the uppercase names?
(basically, edit the commit that introduces these functions instead of adding extra commit on the top).

That would ease the review and keep the `git log` readable :slightly_smiling_face:

lamphamsy:
@slaperche-scality sure, the commits are squashed. Thanks :)

src/simd_128.h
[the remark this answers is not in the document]
slaperche-scality: > I tested that const gives worse performances than define

Are you sure? By how much?

> the use of const will create __static_initialization_and_destruction_0. Maybe it decreases the performances.

I don't think so, I'm almost sure that this function is only called once, before `main` (during the C runtime initialization).

I ran the benchmark (Release mode, SIMD=ON, #samples=1000) on my machine and I haven't got conclusive results that prove a real negative impact from const.

# Benchmark results using define

| FEC        |    m | packet size |     Enc. lat (µs) | Enc. throughput (MB/s) |     Dec. lat (µs) | Dec. throuput (MB/s) |
|:-----------|-----:|------------:|------------------:|-----------------------:|------------------:|---------------------:|
| rs-fnt     |  240 |         256 | 5522.86 ± 222.015 |                2373.26 | 10496.5 ± 467.663 |              78.0454 |
| rs-fnt     |  240 |         512 | 5483.07 ± 239.125 |                2390.48 |   10607 ± 431.391 |              77.2318 |
| rs-fnt     |  240 |        1024 | 5366.83 ± 230.948 |                2442.26 | 10342.7 ±  433.89 |               79.206 |
| rs-fnt     | 1008 |         256 | 23123.1 ± 807.104 |                2267.38 | 50222.8 ± 1436.43 |              16.3113 |
| rs-fnt     | 1008 |         512 | 23857.8 ± 1066.64 |                2197.56 | 47290.8 ± 1941.65 |              17.3226 |
| rs-fnt     | 1008 |        1024 | 27243.1 ± 1149.77 |                1924.48 |   60185 ± 3605.52 |              13.6114 |
| rs-fnt-sys |  240 |         256 | 13804.7 ± 476.495 |                890.131 | 15471.8 ±  559.92 |               52.948 |
| rs-fnt-sys |  240 |         512 | 14075.7 ± 534.422 |                872.994 | 15321.9 ± 575.059 |              53.4661 |
| rs-fnt-sys |  240 |        1024 | 14343.9 ± 650.119 |                856.669 | 15152.5 ± 580.651 |              54.0637 |
| rs-fnt-sys | 1008 |         256 | 65326.7 ± 2384.91 |                790.024 | 70649.7 ± 2187.57 |              11.5952 |
| rs-fnt-sys | 1008 |         512 | 72595.2 ± 2733.07 |                710.923 |   73075 ±    3558 |              11.2104 |
| rs-fnt-sys | 1008 |        1024 | 80948.2 ± 3173.47 |                637.564 | 90715.6 ± 3893.63 |              9.03042 |

# Benchmark results using const

| FEC        |    m | packet size |     Enc. lat (µs) | Enc. throughput (MB/s) |     Dec. lat (µs) | Dec. throuput (MB/s) |
|:-----------|-----:|------------:|------------------:|-----------------------:|------------------:|---------------------:|
| rs-fnt     |  240 |         256 | 5508.43 ± 227.965 |                2379.48 | 10360.4 ± 412.958 |              79.0706 |
| rs-fnt     |  240 |         512 |  5480.7 ± 234.837 |                2391.52 | 10552.5 ± 432.283 |              77.6305 |
| rs-fnt     |  240 |        1024 | 5429.65 ± 281.588 |                   2414 | 10338.2 ± 427.177 |              79.2404 |
| rs-fnt     | 1008 |         256 | 23129.9 ± 907.399 |                2266.71 | 49949.6 ± 1596.99 |              16.4005 |
| rs-fnt     | 1008 |         512 | 24179.1 ± 1458.99 |                2168.35 | 47136.3 ± 3114.64 |              17.3794 |
| rs-fnt     | 1008 |        1024 | 26870.8 ± 1176.66 |                1951.14 |   59786 ± 3434.83 |              13.7022 |
| rs-fnt-sys |  240 |         256 | 13777.5 ± 562.277 |                891.889 |   15263 ± 514.184 |              53.6722 |
| rs-fnt-sys |  240 |         512 | 14071.1 ± 536.787 |                 873.28 | 15303.4 ± 547.433 |              53.5305 |
| rs-fnt-sys |  240 |        1024 | 14326.6 ± 645.201 |                857.702 | 15114.8 ± 674.564 |              54.1986 |
| rs-fnt-sys | 1008 |         256 | 65830.4 ± 2277.08 |                783.978 | 70598.5 ± 2168.34 |              11.6036 |
| rs-fnt-sys | 1008 |         512 | 72133.8 ± 2794.12 |                 715.47 | 73205.2 ± 3214.68 |              11.1905 |
| rs-fnt-sys | 1008 |        1024 | 82156.5 ± 2912.75 |                628.186 | 89188.1 ± 3586.39 |              9.18508 |

# Comparison

## Encoding

| define (MB/s) | const (MB/s) | diff (%) |
|--------------:|-------------:|---------:|
|       2373.26 |      2379.48 |     0.26 |
|       2390.48 |      2391.52 |     0.04 |
|       2442.26 |         2414 |    -1.16 |
|       2267.38 |      2266.71 |    -0.03 |
|       2197.56 |      2168.35 |    -1.33 |
|       1924.48 |      1951.14 |     1.39 |
|       890.131 |      891.889 |      0.2 |
|       872.994 |       873.28 |     0.03 |
|       856.669 |      857.702 |     0.12 |
|       790.024 |      783.978 |    -0.77 |
|       710.923 |       715.47 |     0.64 |
|       637.564 |      628.186 |    -1.47 |


## Decoding

| define (MB/s) | const (MB/s) | diff (%) |
|--------------:|-------------:|---------:|
|       78.0454 |      79.0706 |     1.31 |
|       77.2318 |      77.6305 |     0.52 |
|        79.206 |      79.2404 |     0.04 |
|       16.3113 |      16.4005 |     0.55 |
|       17.3226 |      17.3794 |     0.33 |
|       13.6114 |      13.7022 |     0.67 |
|        52.948 |      53.6722 |     1.37 |
|       53.4661 |      53.5305 |     0.12 |
|       54.0637 |      54.1986 |     0.25 |
|       11.5952 |      11.6036 |     0.07 |
|       11.2104 |      11.1905 |    -0.18 |
|       9.03042 |      9.18508 |     1.71 |

src/fft_2n.cpp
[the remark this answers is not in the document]
slaperche-scality: The new code looks better, it's OK I think.

Good job :+1:

src/simd_fnt.h
slaperche-scality: Makes sense

src/simd_fnt.h
slaperche-scality: There is still one occurrence in `fft_2n.h`

```
% git-grep -i butterly
src/fft_2n.h:413: * Butterly CT on two-layers at a time
```

src/property.h
```diff
@@ -40,6 +40,8 @@
 
 namespace quadiron {
 
+#define OOR_MARK 1
```
slaperche-scality: We could use a `static const int` (or `static constexpr int`)

src/simd_128.h
```diff
@@ -56,32 +56,32 @@ typedef __m128i VecType;
 
 /* ============= Essential Operations for SSE w/ both u16 & u32 ============ */
 
-inline VecType LOAD(VecType* address)
+inline VecType LoadToReg(VecType* address)
```
slaperche-scality: Why the `CamelCase`?

All the other functions are named in `snake_case`?

Also, we should not have a commit "rename things", you should do an interactive rebase and fix the naming directly in the commits that introduce these name (" SIMD 128: essential operations for SSE" and "SIMD 256: essential operations for AVX").

src/simd_basic.h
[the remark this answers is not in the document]
slaperche-scality: The new code is more readable and with less duplication, good job!

src/simd_basic.h
```diff
@@ -0,0 +1,404 @@
...
+ *  add result to correspondent element of `dest`
+ *
+ * @note: 1 < `a` < card - 1
+ */
+template <typename T>
+inline void mul_coef_to_buf(const T a, T* src, T* dest, size_t len, T card)
+{
+    const VecType coef = SET1(a);
+
+    VecType* __restrict _src = reinterpret_cast<VecType*>(src);
```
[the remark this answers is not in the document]
slaperche-scality: Yeah I know what's the role of **restrict** is, my points were:

1. **restrict** is a C99 keyword, not a C++ one, so we have to use `__restrict__` (but after a quick check, `__restrict__` is supported by Clang and GCC whereas `__restrict` is supported by Clang, GCC and Microsoft Visual C++)

=> so you're right, better use `__restrict`.

2. using **restrict** while violating the strict aliasing rules (which we are doing in several place in our codebase, such as #256 (but that's not the only place)) is an **Undefined Behavior**

=> I'm not sure we should use **restrict** except if we are 100% sure that we are not violating strict aliasing rules here.

src/simd_nf4.h
```diff
@@ -162,11 +161,227 @@ inline aint128 pack(__uint128_t a, uint32_t flag)
...
+    __m128i b = _mm_set_epi32(b3, b2, b1, b0);
 
     return m128i_to_uint128(b);
 }
 
+/* ================= Basic operations for NF4 ================= */
+
+#if defined(__AVX2__)
+
+inline VecType CAST_TO_DOUBLE(HalfVecType x)
```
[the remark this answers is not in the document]
slaperche-scality: Good naming (LoadToReg), but you should use `snake_case` (`load_to_reg`) like all the functions.

src/simd_nf4.h
```diff
@@ -162,11 +161,227 @@ inline aint128 pack(__uint128_t a, uint32_t flag)
...
+{
+    VecType res;
+    STORE(&res, MULFULL_MOD((VecType)a, (VecType)b, F4));
+    return (NF4Type)res;
+}
+
+inline void
+add_buf_to_two_bufs_rem(unsigned n, NF4Type* x, NF4Type* x_half, NF4Type* y)
+{
+    // do nothing
```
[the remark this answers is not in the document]
slaperche-scality: Ok, thanks for the explanation

src/simd_128.h
```diff
@@ -81,6 +81,10 @@ inline uint16_t TESTZ(VecType x, VecType y)
 {
     return _mm_testz_si128(x, y);
 }
+inline int is_all_zeros(VecType x)
```
slaperche-scality: This should return a `bool` since it's a predicate (starts with `is_`).

src/simd_basic.h
```diff
@@ -129,17 +129,16 @@ inline VecType MUL_MOD(VecType x, VecType y, T q)
...
     // filter elements of both of a & b = card-1
     const VecType cmp = AND(CMPEQ<T>(x, CARD_M_1(q)), CMPEQ<T>(y, CARD_M_1(q)));
-    res = (q == F3) ? XOR(res, AND(F4_u32, cmp)) : ADD<T>(res, AND(ONE32, cmp));
 
-    const VecType lo =
-        (q == F3) ? BLEND8(ZERO, res, MASK8_LO) : BLEND16(ZERO, res, 0x55);
-    const VecType hi = (q == F3) ? BLEND8(ZERO, SHIFTR(res, 1), MASK8_LO)
-                           : BLEND16(ZERO, SHIFTR(res, 2), 0x55);
-    return SUB_MOD(lo, hi, q);
+    if (is_all_zeros(cmp) == 1) {
```
slaperche-scality: If you apply my comment about returning a `bool` for `is_all_zeros`, then you can get rid of the `== 1` part.

src/simd_nf4.h
```diff
@@ -168,9 +168,15 @@ inline __uint128_t pack(__uint128_t a, uint32_t flag)
 
 #if defined(__AVX2__)
 
-inline VecType CAST_TO_DOUBLE(HalfVecType x)
+inline VecType load_to_reg(HalfVecType x)
```
slaperche-scality: the `_a` and `_b` can probably be const, no?

Also, try to avoid prefixing name with `_`, this is usually reserved for the compiler and internals.

src/simd_128.h
```diff
@@ -40,17 +40,17 @@ typedef __m128i VecType;
 
 /* ============= Constant variable  ============ */
 
-#define F4_u32 _mm_set1_epi32(65537)
-#define F4m1_u32 _mm_set1_epi32(65536)
-#define F3_u32 _mm_set1_epi32(257)
-#define F3m1_u32 _mm_set1_epi32(256)
+#define F4_U32 _mm_set1_epi32(65537)
```
slaperche-scality: Good, but we shouldn't have a patch for renaming: you should fix the naming directly in the patch that introduce them (with an interactive rebase)

src/simd_basic.h
```diff
@@ -37,13 +37,27 @@ namespace quadiron {
...
+}
+template <>
+inline VecType Card<uint32_t>(uint32_t q)
 {
     return (q == F3) ? F3_U32 : F4_U32;
 }
 
 template <typename T>
-inline VecType CardMinusOne(T q)
+inline VecType CardMinusOne(T q);
```
slaperche-scality: ditto (should be `card_minus_one`)

src/simd_basic.h
```diff
@@ -37,13 +37,27 @@ namespace quadiron {
 namespace simd {
 
 template <typename T>
-inline VecType Card(T q)
+inline VecType Card(T q);
+template <>
+inline VecType Card<uint16_t>(uint16_t q)
```
slaperche-scality: Don't use CamelCase for function name (should be `card`)

src/simd_basic.h
```diff
@@ -62,6 +62,19 @@ inline VecType CardMinusOne<uint32_t>(uint32_t q)
     return (q == F3) ? F3_MINUS_ONE_U32 : F4_MINUS_ONE_U32;
 }
 
+template <typename T>
+inline VecType GetLowHalf(VecType x, T q)
```
slaperche-scality: No CamelCase for function name (should be `get_low_half`)

src/simd_basic.h
```diff
@@ -62,6 +62,19 @@ inline VecType CardMinusOne<uint32_t>(uint32_t q)
...
+{
+    return (q == F3) ? BLEND8(ZERO, x, MASK8_LO) : BLEND16(ZERO, x, 0x55);
+}
+
+template <typename T>
+inline VecType GetHighHalf(VecType x, T q)
```
slaperche-scality: No CamelCase for function name (should be `get_high_half`)

src/simd_basic.h
lamphamsy: - [x] rename

src/simd_basic.h
lamphamsy: - [x] rename

src/simd_basic.h
lamphamsy: - [x] rename

src/simd_basic.h
lamphamsy: - [x] rename

src/simd_fnt.h
lamphamsy: - [ ] test diff performance w/ and w/o __restrict

src/simd_256_u16.h
```diff
@@ -36,173 +36,629 @@
 namespace quadiron {
 namespace simd {
 
-/** Perform a%card where a is a addition of two numbers whose elements are
- *  symbols of GF(card) */
-inline m256i mod_after_add(m256i a, aint16 card)
-{
-    const m256i _card = _mm256_set1_epi16(card);
-    const m256i _card_minus_1 = _mm256_set1_epi16(card - 1);
+#define F3_u16 _mm256_set1_epi16(257)
```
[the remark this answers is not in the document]
lamphamsy: These commits implement new algorithms for modular operations. Next commits are based on it to refactor codes :)

src/simd_128.h
lamphamsy: yes, it's better

- [ ] `is_xx` return bool

src/property.h
```diff
@@ -40,6 +40,8 @@
 
 namespace quadiron {
 
+static const unsigned OOR_MARK = 1;
```
slaperche-scality: Could even be a constexpr I think. (just replace `const` by `constexpr`).

src/simd_256_u16.h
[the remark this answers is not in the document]
slaperche-scality: That's not what I'm seeing in the Git history.

- `SIMD 256 u16 & u32: update` add codes in `src/simd_256_u16.h`
- `SIMD 256 u16 u32: remove useless files` delete the WHOLE file (so all the code is removed).
- And finally, ` SIMD 256: essential operations for AVX` reimplements the content but differently than from `SIMD 256 u16 & u32: update`

By looking at the Git history, it seems like only the 2 last commits useful, the first one (this one and its sibling for SSE) can be removed.

And in fact, if I checkout your branch and remove these 2 commits during an interactive rebase, then everything works, with no conflict.

So you should remove them, they are just confusing.

src/property.h
slaperche-scality: Could even be a constexpr I think. (just replace `const` by `constexpr`).

`constexpr` can be used in more context than `const`, so they should be preferred when possible ;)

src/simd_basic.h
```diff
@@ -145,7 +145,7 @@ inline void ADD_PROPS(
 {
     const VecType b = CMPEQ32(threshold, symb);
     const VecType c = AND(mask, b);
-    MaskIntType d = MVMSK8(c);
+    auto d = MVMSK8(c);
```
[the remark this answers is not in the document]
slaperche-scality: Ah, I get it. It depends on the instruction set, not on the type T.

src/simd_256_u16.h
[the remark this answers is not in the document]
lamphamsy: you're right. I will remove them

- [ ] remove useless commits

slaperche-scality:
Nice squashing, the PR looks better now.

I think we can still squash some:

- "SIMD 256 u16 u32: remove useless files" + "SIMD 256 u16 u32: remove useless files" =  "SIMD{128,256}/{u16,u32}: remove useless files"
- "FFT_2n.h: compute indices for SIMD" + "FFT_2n.cpp: compute indices for SIMD" = "FFT_2n: compute indices for SIMD"
- "FEC RS FNT: simd indices as member variables" + "FEC Vectorisation: use FNT's simd indices"

src/simd_256.h
```diff
@@ -57,28 +57,26 @@ typedef __m128i HalfVecType;
 // with static storage duration may throw an exception that cannot be caught
 
 // NOLINTNEXTLINE(cert-err58-cpp)
-const VecType F4_U32 = _mm256_set1_epi32(65537);
+const VecType vec32_f4 = _mm256_set1_epi32(65537);
```
slaperche-scality: I think you should keep the UPPERCASE naming.

They're not #define anymore, but they are still global constants and as entities with a global scope they should be in UPPERCASE.

lamphamsy:
@slaperche-scality thank you 👍

PR #255, opened 2018-10-16
Bootstrap the new SIMD module

slaperche-scality:
This PR just lay the foundation for the new SIMD module, the biggest part will come in the next PR.

That being said, here we defines the basic constants and helpers, as well as a custom allocator.

### DIFF
--- a/src/simd/allocator.h
+++ b/src/simd/allocator.h
@@ -1,0 +1,165 @@
+/*
+ * Copyright 2017-2018 Scality
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/** @file allocator.h
+ *
+ * Provide a custom memory allocator for SIMD.
+ *
+ * This allocator always returns memory that is suitably aligned to be loaded
+ * efficiently by a Register object.
+ */
+
+#ifndef __QUAD_SIMD_SIMD_ALLOCATOR_H__
+#define __QUAD_SIMD_SIMD_ALLOCATOR_H__
+
+#include <cassert>
+#include <cstdint>
+#include <limits>
+
+#include "simd/definitions.h"
+
+namespace quadiron {
+namespace simd {
+
+/// Check that the given address is properly aligned.
+template <typename T>
+inline bool addr_is_aligned(const T* addr)
+{
+    // Without SIMD, there is no specific alignment constraint.
+    if (INSTRUCTION_SET == InstructionSet::NONE) {
+        return true;
+    }
+    const std::uintptr_t address = reinterpret_cast<std::uintptr_t>(addr);
+    return (address & (ALIGNMENT - 1)) == 0;
+}
+
+/** Custom allocator to take advantage of SIMD processing.
+ *
+ * This allocator always return memory that is suitably aligned for the current
+ * SIMD instruction set. Thanks to this property, you can safely use the aligned
+ * load from the Register class in order to increase performance.
+ */
+template <typename T>
+class AlignedAllocator {
+  public:
+    using value_type = T;
+
+    AlignedAllocator() noexcept {}
+    // No state, => nothing to copy.
+    template <class U>
+    AlignedAllocator(AlignedAllocator<U> const& other) noexcept
+    {
+    }
+
+    value_type* allocate(std::size_t count)
+    {
+        // Guard against overflow!
+        if (count > max_size()) {
+            throw std::bad_alloc();
+        }
+
+        // No SIMD: default allocator is good enough!
+        if (INSTRUCTION_SET == InstructionSet::NONE) {
+            return static_cast<value_type*>(
+                ::operator new(count * sizeof(value_type)));
+        }
+
+        // Overallocate just enough to have room for alignment adjustment.
+        const std::size_t size = count * sizeof(value_type) + ALIGNMENT;
+        unsigned char* ptr = static_cast<unsigned char*>(::operator new(size));
+
+        // Align the allocated memory.
+        const std::uintptr_t address = reinterpret_cast<std::uintptr_t>(ptr);
+        const unsigned offset = ALIGNMENT - (address % ALIGNMENT);
+        assert(offset >= 1); // We need a byte to store the offset itself.
+        unsigned char* aligned_ptr = ptr + offset;
+
+        // Store the offset just before the aligned memory.
+        assert(offset <= std::numeric_limits<unsigned char>::max());
+        *(aligned_ptr - 1) = static_cast<unsigned char>(offset);
+
+        // Return the aligned pointer.
+        //
+        // Clang analyser think that we leak `ptr`, whereas we can re-compute
+        // it from `aligned_ptr` and free it in `deallocate`.
+        // NOLINTNEXTLINE(clang-analyzer-cplusplus.NewDeleteLeaks)
+        return reinterpret_cast<value_type*>(aligned_ptr);
+    }
+
+    void deallocate(value_type* ptr, std::size_t count) noexcept
+    {
+        // No SIMD: default allocator is good enough!
+        if (INSTRUCTION_SET == InstructionSet::NONE) {
+            ::operator delete(ptr);
+            return;
+        }
+
+        if (ptr == nullptr) {
+            return;
+        }
+        // Respect strict aliasing rules: read through a character type.
+        unsigned char* raw = reinterpret_cast<unsigned char*>(ptr);
+        // Get the alignment offset stored just before the aligned pointer.
+        const unsigned offset = *(raw - 1);
+        ::operator delete(raw - offset);
+    }
+
+    std::size_t max_size() const noexcept
+    {
+        const std::size_t max_size = std::numeric_limits<std::size_t>::max();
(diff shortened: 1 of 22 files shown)


src/simd/allocator.h
```diff
@@ -0,0 +1,165 @@
...
+    value_type* allocate(std::size_t count)
+    {
+        // Guard against overflow!
+        if (count > max_size()) {
+            throw std::bad_alloc();
+        }
+
+        // No SIMD: default allocator is good enough!
+        if (INSTRUCTION_SET == InstructionSet::NONE) {
+            return static_cast<value_type*>(
```
lamphamsy: Why don't we use `new T[count]`?

src/simd/allocator.h
```diff
@@ -0,0 +1,165 @@
...
+
+        // Return the aligned pointer.
+        //
+        // Clang analyser think that we leak `ptr`, whereas we can re-compute
+        // it from `aligned_ptr` and free it in `deallocate`.
+        // NOLINTNEXTLINE(clang-analyzer-cplusplus.NewDeleteLeaks)
+        return reinterpret_cast<value_type*>(aligned_ptr);
+    }
+
+    void deallocate(value_type* ptr, std::size_t count) noexcept
```
lamphamsy: the 2nd argument is useless, isn't it?

src/simd/definitions.h
```diff
@@ -0,0 +1,124 @@
...
+    SSE,  ///< SSE4.1
+    AVX,  ///< AVX2
+};
+
+// Definitions for Intel AVX-256 {{{
+
+// We required AVX2 because we relies on some instructions (such as
+// `_mm256_add_epi16` and others) that aren't available in the first version of
+// AVX.
+#if defined(__AVX__) && defined(__AVX2__)
```
lamphamsy: Is `defined(__AVX2__)` enough?

src/simd/definitions.h
```diff
@@ -0,0 +1,124 @@
...
+
+using RegisterType = __m256;
```
lamphamsy: We have `MaskType`, so do we need this type?

src/simd/definitions.h
```diff
@@ -0,0 +1,124 @@
...
+
+#elif defined(__SSE__) && defined(__SSE4_1__)
+
+using RegisterType = __m128;
+using MaskType = __m128i;
+
+static constexpr InstructionSet INSTRUCTION_SET = InstructionSet::SSE;
+
+// }}}
+// Definitions for scalar fallback {{{
```
lamphamsy: Do we need definitions for non-supported SIMD?

src/simd_nf4.h
```diff
@@ -52,19 +54,18 @@ static inline aint128 m128i_to_uint128(m128i v)
 inline aint128 expand16(uint16_t* arr, int n)
 {
     // since n <= 4
-    uint16_t _arr[4] __attribute__((aligned(ALIGN_SIZE))) = {0, 0, 0, 0};
+    uint16_t _arr[4] __attribute__((aligned(simd::ALIGNMENT))) = {0, 0, 0, 0};
```
lamphamsy: It is in `simd` namespace so `simd::` is not necessary

src/simd_nf4.h
```diff
@@ -117,8 +117,7 @@ inline void unpack(__uint128_t a, GroupedValues<__uint128_t>& b, int n)
     const uint32_t flag =
         ai[1] | (!!ai[3] << 1u) | (!!ai[5] << 2u) | (!!ai[7] << 3u);
 
-    m128i val = _mm_set_epi64(
-        _mm_setzero_si64(), _mm_set_pi16(ai[6], ai[4], ai[2], ai[0]));
+    m128i val = _mm_set_epi16(0, 0, 0, 0, ai[6], ai[4], ai[2], ai[0]);
```
lamphamsy: 👍

src/simd/allocator.h
slaperche-scality: There is a difference between those two.
- the form I used is the "raw" `operator new` (aka the global `operator new`), which behaves like `malloc` in C: it returns a block of uninitialized storage (no object has been constructed within that storage, thus it has no dynamic type).
- the form you suggest is the most common and combines two steps: first the memory is allocated (using the "raw" operator) and then it construct an object of type T (hence this piece of memory has a dynamic type T).

Now, does it matter? After all, in our very specific case we mainly allocate native integer types, which doesn't have a constructor so it should make no difference, right?
In fact it does matter, because of strict aliasing rules: in the previous version of our allocation function (which uses `new uint8_t[len]`) we weren't allocation a piece of memory (like `malloc` does in C), we were creating an array of `uint8_t` from then on we were subject to the strict aliasing rule (the line `T* data = reinterpret_cast<T*>(ptr);` was breaking this rule and thus was an undefined behavior).

The new version, that uses the "raw" `operator new`, return untyped memory (no object are constructed) and thus isn't subject to this rule (because no dynamic type was assigned) until we construct an object either by using placement new (which is another version of the `operator new`), or by calling the method `construct` of the allocator (which usually relies on placement new).

If you wan a quick overview of the different facet of `new` you can read [this blog post](https://eli.thegreenplace.net/2011/02/17/the-many-faces-of-operator-new-in-c).

See also [this](https://stackoverflow.com/a/2697929) and [this](https://stackoverflow.com/a/36021642) which may be more readable than the C++ standard.

src/simd/allocator.h
slaperche-scality: It's unused in my implementation, yes.
That being said it's specified like this (see [the specification of the method](https://en.cppreference.com/w/cpp/memory/allocator/deallocate)), and it can be used some allocator to improve the performance (example [here](https://github.com/jemalloc/jemalloc/issues/28)).

src/simd/definitions.h
slaperche-scality: Yes, it's enough but I prefer to be explicit: here we want AVX **BUT** because of QuadIron uses of SIMD we required AVX2.

src/simd/definitions.h
slaperche-scality: `MaskType` has a different type than `RegisterType`, and in my future PR we will have two different classes (`Register` and `Mask`).
A mask and a register have different interface and constraint, like a mask must be an integer, a register can hold floating-point value (not in our case here in QuadIron, but still I think it's cleaner to separate them: anyway the conversion/cast between those two types are free (no instruction are generated)).

src/simd/definitions.h
slaperche-scality: Yeah, because my idea to reduce code duplication is to try, as much as possible, to have a single code path.

Currently we have many code duplication with:
- `f1` that implement function without SIMD
- `f2.1` which implement function with SSE on 16-bit
- `f2.2` which implement function with SSE on 32-bit
- `f3.1` which implement function with AVX on 16-bit
- `f3.2` which implement function with AVX on 32-bit

So basically, 5 functions (and we don't support yet NEON and AVX256, nor speedup on 64-bit) that are doing the same thing.

If we could have a generic `Register` class with the necessary methods we could write code that could look like

```c++
for (unsigned i = 0; i < data.size(); i += simd::countof<T>()) {
    Register<T> r1(&data[i]);
    Register<T> r2(&coef[i]);

    Register<T> out(&output[i]);
    out = r1 + r2;
}
```

And then, if it's compiled with SIMD the constructor would call the correct ` _mm*_load*` instrisic, the `operator+` would be implemented by using the correct `_mm*_add*` instrinsics, …
And if we compile without SIMD, then `Register` would simply iterate element by element, using native multiplication, …

That way, we could write a single function and still benefit from SIMD acceleration.
All the specific would be hidden in the `Register` class (using template specialization) and our "main" code (the FEC) would be more readable.

I don't know yet how the final code will look like, I'm still working on the implementation, but that's the goad of this whole "rework the SIMD API".

src/simd_nf4.h
slaperche-scality: Good catch! I'll fix this :smiley:

src/simd/allocator.h
lamphamsy: Okay. So does `operator new` allocates a contiguous piece of memory?

src/simd/allocator.h
lamphamsy: so are we going to use this variable?

src/simd/definitions.h
lamphamsy: So I think we could simplify it, also the comment as we need AVX2

src/simd/definitions.h
lamphamsy: It's still not clear for me :)
I think we work only for integers, not floating-point variables

src/simd/definitions.h
lamphamsy: Yes, it looks good. It looks similar what https://github.com/QuantStack/xsimd does
BTW, in my PR #254, all 5 functions are combined in a templated one. There are still only wrappers of SIMD operations and basic operations that depend on register type and 16/32-bit packed elements

src/simd/allocator.h
slaperche-scality: Yes, of course `new` return contiguous memory.
Otherwise, how the language would work (no array, no std::vector, …)?

src/simd/allocator.h
slaperche-scality: No, we won't use it.
But it has to be here anyway, it's part of the allocator interface: if you don't have it then you're not implementing an allocator.

That being said, it's not an issue: the standard says "if you pass another value than the one used to allocate the behavior is undefined", well "undefined" can be anything including "it works" and since we are implementing it here, we get to decide how the code behave: and as you can see, you can give any value you want it will works (because we don't use it), nevertheless it has to be there to fullfill the requirement.

src/simd/definitions.h
slaperche-scality: We could, but I think explicit is better than implicit (and removing __AVX__ doesn't really make the code more or less readable)

src/simd/definitions.h
slaperche-scality: Maybe it will be clearer in the next PRs, when I'll actuelly use it :)

>  I think we work only for integers, not floating-point variables

Yes, in our very specific case we only use integers (for now), but it doesn't hurt to have distinct types for distinct uses (multiplying two mask together doesn't make sense, so you can implement `operator *` for RegisterType and not for `MaskType`, etc).

src/simd/definitions.h
slaperche-scality: I don't know this lib specifically, but yeah I think it's a pretty standard approach.

> BTW, in my PR #254, all 5 functions are combined in a templated one.

On one hand, it's good to hear, on the other hand I'm not sure it's a good idea to have two concurrent PR reworking the SIMD :S
I hope we won't conflict too much…

src/simd/allocator.h
lamphamsy: Okay, so to benefit this advantage, should we use allocation size, e.g. `size = count * sizeof(value_type) + ALIGNMENT` for SIMD, in `::operator delete()` at line 120 & 131?

src/simd/allocator.h
lamphamsy: yes, okay :)

src/simd/definitions.h
lamphamsy: yes, I'd like to see its use in next PRs

src/simd/allocator.h
slaperche-scality: Ha indeed, since C++14 the "standard" `delete` also accept a size.

I can make this change (but then, giving a random size to `deallocate` will really trigger an undefined behavior!).

src/simd/definitions.h
lamphamsy: I think there are only small conflicts. Your PR is on SIMD definitions. In PR #254, I re-implement all modular operations and vectorized functions for Radix2 FFT. So I re-organize these codes to remove duplicated functions.

src/simd/allocator.h
slaperche-scality: Hum, the `void operator delete  ( void* ptr, std::size_t sz );` seems not supported by Clang 6 (but works with GCC 8.2), so I think I'll revert it because we can't lost the ability to compile with Clang for an hypothetical performance improvement.

lamphamsy: Okay, so we don't do that. Anyway, we do no benefit advantage of sized deallocation, don't we?

src/simd/definitions.h
```diff
@@ -0,0 +1,124 @@
...
+
+#else
+
+// Use The Widest Integer Type Available.
+#if __LP64__ == 1
+using RegisterType = uint64_t;
```
lamphamsy: I have a question for this case where SIMD is not supported. If we work of `FEC<uint32_t>`, do we accelerate operations by working on `Register` of type `uint64_t`? i.e. a pack of 2 elements?

slaperche-scality: At first, I wanted to do that, but it may be tricky to do correctly due to the strict aliasing constraints so for now no: items will be processed one by one.

We may try to do that later though (if it's possible to do correctly AND efficiently (I may have some idea, but it may relies on how clever the compiler is)).

src/simd/allocator.h
slaperche-scality: Anyway, I think the standard allocator from the libc/libc++ doesn't take advantage of it yet so yeah: no benefit for now I think (but we still have to keep the prototype).

PR #258, opened 2018-10-21
Add more compilation flags

slaperche-scality:
In order to keep a good level of code quality, this PR introduces makes Clang and GCC stricter.

The set of flags are the ones that were used on the DCSS project.

The next step will be to enable more lint check in clang-tidy, but that PR is already big enough as it is.

Also, before moving onto the linter, there are still two more warnings that I want to enable: `-Wconversion` and `-Wshadow`.
Those will require a litlte bit more rework so they will be done in a dedicated PR:
- `-Wconversion`: it's probably wise to tackle https://github.com/scality/quadiron/issues/243 first
- `-Wshadow` will require a bit of wide renaming (but in the end will improve the code readability) and I'll probably rework the visibility of some fields on the way (too many public fields in some of our classes right now).


slaperche-scality:
@lamphamsy It would be great if you could review the changes in the commit "fix Doxygen documentation (flagged by `-Wdocumentation`)" carefully.

Most of the errors flagged by `-Wdocumentation` were about missing description for function parameters.
I fixed those by either:
- adding a description when I clearly understood what's the role of the parameter
- adding a not-so-useful-but-not-wrong description when it wasn't clear to me
- removing the documentation of the parameters

If you have suggestions for those, I would be more than happy to update this commit with better descriptions/documentation :slightly_smiling_face:

src/gf_ring.h
```diff
@@ -859,34 +820,27 @@ T RingModN<T>::get_primitive_root() const
     return this->root;
 }
 
-/**
- * find smallest number is
- *  - at least n
- *  - divisible by (q-1)
- *
- * @param n
+/** Find smallest number that is greater than `n` and divisible by `q-1`.
```
lamphamsy: `is greater than` => `is at least`

src/gf_ring.h
```diff
@@ -859,34 +820,27 @@ T RingModN<T>::get_primitive_root() const
...
 }
 
-/**
- * find smallest number is
- *  - highly composited
- *  - at least n
- *  - divisible by (q-1)
- *
- * @param n
+/** Find smallest number that is highly composite, greater than `n` and
```
lamphamsy: `greater than` => `at least`

src/vec_vector.h
```diff
@@ -68,9 +68,15 @@ template <typename T>
 class Vector {
   public:
     const gf::RingModN<T>* rn;
+
     Vector(const gf::RingModN<T>& rn, int n, T* mem = nullptr, int mem_len = 0);
     Vector(const gf::RingModN<T>& rn, std::initializer_list<T> values);
+    Vector(const Vector& other);
+    Vector(Vector&& other) noexcept;
```
lamphamsy: I see that these two new constructors are used only for fft_utest.cp. I think it's better to fix the issue in the tests instead of modifying vec_vector.h

src/gf_nf4.h
```diff
@@ -205,127 +214,120 @@ inline T NF4<T>::neg(T a) const
 template <typename T>
 inline T NF4<T>::add(T a, T b) const
 {
-    uint32_t arr[this->n];
-
-    arr[0] = (narrow_cast<uint32_t>(a) + narrow_cast<uint32_t>(b)) % 65537;
+    scratch32[0] =
```
lamphamsy: As we discussed, `scratch16` and `scratch32` could lead to issues if this operation is executed in multi-threads.
Maybe a note about it is useful for now.

src/fft_add.h
```diff
@@ -505,24 +489,19 @@ void Additive<T>::_taylor_expand_t2(
...
+        _taylor_expand_t2(_g1, n - deg0, k - 1, s_deg + deg2);
     }
 }
 
-/*
- * This function compute f(x) from its taylor expansion
- *  f(x) = sum_i (gi0 + gi1 * x) * (x^2 - x)^i
+/** This function compute f(x) from its taylor expansion.
+ *
+ * f(x) = sum_i (gi0 + gi1 * x) * (x^2 - x)^i
```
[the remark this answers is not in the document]
slaperche-scality: Done

src/gf_ring.h
slaperche-scality: Done

src/gf_ring.h
slaperche-scality: Done

src/vec_vector.h
slaperche-scality: The thing is, there is no problem with the tests but there is a problem of bad design with the Vector class here.

The Vector class owns a resource, and to properly handle this we have provided a user-defined destructor. Once you do that, you have to follow the [Rule of Three](https://en.cppreference.com/w/cpp/language/rule_of_three) (Rule of Five since C++11).

In general, when you have a class that manage some resources you have the choice to consider you class either as:
- a value-semantic (that's the case for vector): you have to follow the Rule of 3/5
- an entity-semantic: you should forbid the copy (and maybe implement the moves).

The fix you suggest would be to use `=` as before but that's bad for two reasons:
- this doesn't fix the flaw in the class's design
- using the copy assignment instead of the copy constructor is slower (not like we care in the tests, but that's still ugly)

src/gf_nf4.h
slaperche-scality: The thing is, most of our class are not MT-safe (e.g: you can't call the method `fft` of `fft::Additive` in multiple thread => we will have a data race on the `mem` attribute, same for the `encode` of `fec::RsFnt`, and so on…)

I could add a comment for this one, but then for lot of other classes as well (I can do it, but in a dedicated PR).

The thing is, most of our class are not MT-safe at the instance level (you can do MT if each thread create its own object, but reusing an object accross multiple threads is not supported by QuadIron).

src/vec_vector.h
lamphamsy: Oh you're right 👍

src/gf_nf4.h
lamphamsy: okay, a dedicated PR might be a good for it

PR #262, opened 2018-10-30
Enhance FFT Radix2 & vectorisation part

lamphamsy:
The branch is the target of the two PRs:

- [x]  Part1: #253 
    - Radix-2 FFT enhancement by working on two layers per iteration

- [x] Part2: #254 
    - Vectorize operations for Radix-2 FFT
    - Re-work also vectorization part to avoid duplication codes

The PR is active when the PRs are merged.


src/simd_fnt.h
```diff
@@ -473,9 +473,8 @@ inline void encode_post_process(
     T threshold,
     size_t vecs_nb)
 {
-    const unsigned element_size = sizeof(T);
     const unsigned vec_size = countof<T>();
-    const T max = 1 << (element_size * 8 - 1);
+    const T max = 1U << (sizeof(T) * 8 - 1);
```
slaperche-scality: minor: could use CHAR_BIT instead of 8 (to be 100% portable, even so 8 is probably enough for the platform we target).

lamphamsy:
@slaperche-scality thanks a lot!

PR #263, opened 2018-10-31
Merging SIMD parts

lamphamsy:
It merges SIMD parts together.

It waits for

- [ ] #262


slaperche-scality:
I think it's too early to merge things now.

I was developing my stuff inside a "simd" subdirectory to avoid conflict with the current SIMD implementation.
Since it's still a work in progress, I think we should keep them separate for now.

PR #277, opened 2019-02-20
Cleanup/fft

vrancurel:
set of cleanups  for release

### DIFF
--- a/src/arith.h
+++ b/src/arith.h
@@ -77,7 +77,7 @@ int jacobi(SignedDoubleSizeVal<T> n, SignedDoubleSizeVal<T> m);
 template <typename T>
 bool solovay_strassen1(T a, T n);
 template <typename T>
-bool solovay_strassen(T n);
+bool solovay_strassen(T n, T (*rand_func)(T n));
 template <typename T>
 bool is_prime(T n);
 template <typename T>
@@ -435,24 +435,27 @@ bool solovay_strassen1(T a, T n)
 {
     const T exponent = (n - 1) / 2;
     const T res = exp_mod(a, exponent, n);
-    int j = jacobi(a, n);
+    int j = jacobi<T>(a, n);
+    T _res;
     if (j < 0) {
-        res = res - n;
+        _res = res - n;
+    } else {
+        _res = res;
     }
-    return j == res;
+    return j == _res;
 }
 
 /** Perform the Solovay–Strassen primality test.
  *
  * @param n check if `n` is prime
+ * @param rand_func return a random number between 0 and n-1
  * @return true if `n` is (probably) prime else false.
  */
 template <typename T>
-bool solovay_strassen(T n)
+bool solovay_strassen(T n, T (*rand_func)(T n))
 {
-    std::uniform_int_distribution<uint32_t> dis(1, n - 1);
     for (int i = 0; i < 100; i++) {
-        T a = dis(prng());
+        T a = rand_func(n);
         if (!solovay_strassen1(a, n))
             return false;
     }
@@ -499,7 +502,7 @@ std::vector<T> factor_distinct_prime(T nb)
         nb = nb / 2;
     }
     // `nb` must be odd at this point => we can skip one element.
-    assert(nb & 1);
+    assert((nb & 1) == 1);
     for (T i = 3; i <= sqrt<T>(nb); i = i + 2) {
         // While `i` divides `nb`, get `i` and divide `nb`.
         while (nb % i == 0) {
@@ -563,7 +566,7 @@ void factor_prime(T nb, std::vector<T>* primes, std::vector<int>* exponent)
         occurence = 0;
     }
     // `nb` must be odd at this point => we can skip one element.
-    assert(nb & 1);
+    assert((nb & 1) == 1);
     for (T i = 3; i <= sqrt<T>(nb); i = i + 2) {
         // While `i` divides `nb`, get `i` and divide `nb`.
         while (nb % i == 0) {
--- a/src/simd/simd.h
+++ b/src/simd/simd.h
@@ -31,8 +31,8 @@
 #ifndef __QUAD_SIMD_SIMD_H__
 #define __QUAD_SIMD_SIMD_H__
 
-#include "simd/allocator.h"
-#include "simd/definitions.h"
+#include "allocator.h"
+#include "definitions.h"
 
 namespace quadiron {
 
--- a/src/vec_vector.h
+++ b/src/vec_vector.h
@@ -46,7 +46,7 @@
 namespace quadiron {
 
 template <typename T>
-class Polynomial;
+class Poly;
 
 /** Low-level wrappers/helpers around memory buffer. */
 namespace vec {
@@ -100,7 +100,8 @@ class Vector {
     void copy(Vector<T>* v, int n, int dest_offset, int src_offset);
     friend bool operator==<T>(const Vector<T>& lhs, const Vector<T>& rhs);
     virtual void neg();
-    void to_poly(Polynomial<T>* poly);
+    void to_poly(Poly<T>* poly);
+    void rand();
     virtual void dump(void) const;
     void sort();
     void swap(unsigned i, unsigned j);
@@ -438,7 +439,7 @@ void Vector<T>::neg()
 }
 
 template <typename T>
-void Vector<T>::to_poly(Polynomial<T>* poly)
+void Vector<T>::to_poly(Poly<T>* poly)
 {
     poly->clear();
     for (int i = 0; i < this->n; i++) {
@@ -457,6 +458,14 @@ void Vector<T>::swap(unsigned i, unsigned j)
 {
     using std::swap;
     swap(mem[i], mem[j]);
+}
+
+template <typename T>
+void Vector<T>::rand()
+{
+    for (int i = 0; i < this->n; i++) {
+        set(i, rn->rand());
+    }
 }
 
 template <typename T>
--- a/test/fft_utest.cpp
+++ b/test/fft_utest.cpp
@@ -33,13 +33,18 @@
 #include "fft_add.h"
 #include "fft_ct.h"
 #include "fft_gt.h"
+#include "fft_large.h"
 #include "fft_naive.h"
 #include "fft_single.h"
 #include "gf_bin_ext.h"
 #include "gf_prime.h"
+#include "misc.h"
+#include "vec_poly.h"
 
 namespace fft = quadiron::fft;
 namespace gf = quadiron::gf;
+namespace arith = quadiron::arith;
+namespace vec = quadiron::vec;
 
 template <typename T>
 class FftTest : public ::testing::Test {
@@ -47,10 +52,10 @@ class FftTest : public ::testing::Test {
     const unsigned q = 65537;
     const unsigned code_len = 32;
 
-    quadiron::vec::Vector<T>
+    vec::Vector<T>
     random_vec(const gf::Field<T>& gf, int size, unsigned to_init)
     {
-        quadiron::vec::Vector<T> vec(gf, size);
+        vec::Vector<T> vec(gf, size);
         vec.zero_fill();
 
         for (unsigned i = 0; i < to_init; i++) {
@@ -67,16 +72,16 @@ class FftTest : public ::testing::Test {
             const int t = 2 + gf.rand() % (fft->get_n() - 1);
             const int n = t + gf.rand() % (fft->get_n() - t + 1);
 
-            quadiron::vec::Vector<T> v1(this->random_vec(gf, n, n));
+            vec::Vector<T> v1(this->random_vec(gf, n, n));
 
             int m = n / t;
             while (m * t < n) {
                 m++;
             }
-            quadiron::vec::Vector<T> v2(gf, t * m);
+            vec::Vector<T> v2(gf, t * m);
 
             fft->taylor_expand(v2, v1, n, t);
-            quadiron::vec::Vector<T> _v1(gf, n);
+            vec::Vector<T> _v1(gf, n);
             fft->inv_taylor_expand(_v1, v2, t);
             ASSERT_EQ(_v1, v1);
         }
@@ -87,10 +92,10 @@ class FftTest : public ::testing::Test {
     {
         for (int i = 0; i < 1000; i++) {
             const int n = fft->get_n();
-            quadiron::vec::Vector<T> v1(this->random_vec(gf, n, n));
+            vec::Vector<T> v1(this->random_vec(gf, n, n));
 
             fft->taylor_expand_t2(v1, n, true);
-            quadiron::vec::Vector<T> _v1(gf, n);
+            vec::Vector<T> _v1(gf, n);
             fft->inv_taylor_expand_t2(_v1);
             ASSERT_EQ(_v1, v1);
         }
@@ -101,11 +106,10 @@ class FftTest : public ::testing::Test {
         fft::FourierTransform<T>* fft,
         int n_data)
     {
-        quadiron::vec::Vector<T> _v(gf, fft->get_n());
-        quadiron::vec::Vector<T> v2(gf, fft->get_n());
+        vec::Vector<T> _v(gf, fft->get_n());
+        vec::Vector<T> v2(gf, fft->get_n());
         for (int j = 0; j < 1000; j++) {
-            quadiron::vec::Vector<T> v(
-                this->random_vec(gf, fft->get_n(), n_data));
+            vec::Vector<T> v(this->random_vec(gf, fft->get_n(), n_data));
 
             fft->fft(_v, v);
             fft->ifft(v2, _v);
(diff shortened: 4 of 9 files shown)


src/arith.h
```diff
@@ -435,24 +435,29 @@ bool solovay_strassen1(T a, T n)
 {
     const T exponent = (n - 1) / 2;
     const T res = exp_mod(a, exponent, n);
-    int j = jacobi(a, n);
+    SignedDoubleSizeVal<T> _a = a;
+    SignedDoubleSizeVal<T> _n = n;
+    int j = jacobi<T>(_a, _n);
```
slaperche-scality: Why do you need `_a` and `_n`?

The input type of `jacoby` are already `SignedDoubleSizeVal<T>` so what was the problem?

src/vec_vector.h
```diff
@@ -46,7 +46,7 @@
 namespace quadiron {
 
 template <typename T>
-class Polynomial;
+class Poly;
```
slaperche-scality: Why shorten the name (and making it less explicit)?

`Polynomial` is not that long, especially with autocomplete, and its easier to grep for.

test/fft_utest.cpp
```diff
@@ -483,3 +549,244 @@ TYPED_TEST(FftTest, TestFftGf2n) // NOLINT
...
+        }
+    }
+
+    // std::cout << "z=" << z << "\n";
+    ASSERT_EQ(
+        z,
+        std::string("921490395895362412399910100421159322")
+            + "712298564831565484737491129935640058571771024");
+
+    delete sfXY;
```
slaperche-scality: use `unique_ptr` to avoid these `delete`.

test/fft_utest.cpp
```diff
@@ -483,3 +549,244 @@ TYPED_TEST(FftTest, TestFftGf2n) // NOLINT
...
+    std::string z("0");
+    for (int i = 0; i <= fft.get_n() - 1; i++) {
+        std::string b(std::to_string(sfXY->get(i)));
+        if (b != std::string("0")) {
+            b.append(i, '0');
+            z = this->big_num_add(z, b);
+        }
+    }
+
+    // std::cout << "z=" << z << "\n";
```
slaperche-scality: dead code, remove

src/simd/simd.h
```diff
@@ -31,8 +31,8 @@
 #ifndef __QUAD_SIMD_SIMD_H__
 #define __QUAD_SIMD_SIMD_H__
 
-#include "simd/allocator.h"
-#include "simd/definitions.h"
+#include "allocator.h"
```
slaperche-scality: That's weird, then something is probably wrong in the install, not in the include directive.

test/fft_utest.cpp
```diff
@@ -140,6 +144,68 @@ class FftTest : public ::testing::Test {
...
+     * @param num
+     *
+     * @return
+     */
+    quadiron::vec::Vector<T>*
+    _convert_string2vec(const gf::Field<T>& gf, int n, char num[])
+    {
+        int i;
+        quadiron::vec::Vector<T>* vec = new quadiron::vec::Vector<T>(gf, n);
+        int len = strlen(num);
```
slaperche-scality: Should use std::string instead of `char[]`?

test/fft_utest.cpp
```diff
@@ -140,6 +144,68 @@ class FftTest : public ::testing::Test {
             ASSERT_EQ(ifft_1, v);
         }
     }
+
+    /**
```
slaperche-scality: Squash the two lines, should be `/** Convert a number into a vector of digits padded with zeros.`

test/fft_utest.cpp
```diff
@@ -140,6 +144,68 @@ class FftTest : public ::testing::Test {
...
+        for (i = 0; i < len; i++) {
+            vec->set(i, num[len - i - 1] - '0');
+        }
+        for (; i < n; i++) {
+            vec->set(i, 0);
+        }
+        return vec;
+    }
+
+    /** compute sum of 2 large numbers in string representation
```
slaperche-scality: `/** Compute sum of 2 large numbers in string representation.`

(proper sentence: add capital and period)

test/fft_utest.cpp
```diff
@@ -140,6 +144,68 @@ class FftTest : public ::testing::Test {
...
+     *
+     * @param str1 first string
+     * @param str2 second string
+     *
+     * @return the sum string
+     */
+    std::string big_num_add(std::string str1, std::string str2)
```
slaperche-scality: `const std::string&` instead of pass-by-copy.

test/fft_utest.cpp
```diff
@@ -140,6 +144,68 @@ class FftTest : public ::testing::Test {
...
+    {
+        if (str1.length() > str2.length())
```
slaperche-scality:
```
if (str1.length() > str2.length()) {
    std::swap(str1, str2);
}
```

test/fft_utest.cpp
```diff
@@ -140,6 +144,68 @@ class FftTest : public ::testing::Test {
...
+            std::swap(str1, str2);
+
+        std::string str = "";
```
slaperche-scality: the default constructor already does that, so you can simply write: `std::string str;`.

Also, since you know the final size beforehand, you could preallocate the space to avoid realloc when `push_back`, by using `reserve()`.

test/fft_utest.cpp
```diff
@@ -140,6 +144,68 @@ class FftTest : public ::testing::Test {
...
+
+        int n1 = str1.length(), n2 = str2.length();
```
slaperche-scality: One declaration per line + `length` probably returns `size_t` so:

```
size_t n1 = str1.length();
size_t n2 = str2.length();
```

But anyway, getting the length is O(1) (not O(n) like `strlen`), you don't can use them in the `for` loop directly (the compiler will optimize).

test/fft_utest.cpp
```diff
@@ -140,6 +144,68 @@ class FftTest : public ::testing::Test {
...
+
+        std::reverse(str1.begin(), str1.end());
```
slaperche-scality: Instead of reversing the string you  could simply use reverse iterator to walk though the strings (using `rbegin` and `rend`).

test/fft_utest.cpp
```diff
@@ -140,6 +144,68 @@ class FftTest : public ::testing::Test {
...
+            str.push_back(sum % 10 + '0');
+            carry = sum / 10;
+        }
+
+        for (int i = n1; i < n2; i++) {
+            int sum = ((str2[i] - '0') + carry);
+            str.push_back(sum % 10 + '0');
+            carry = sum / 10;
+        }
+
```
slaperche-scality:
```
if (carry) {
            str.push_back(carry + '0');
}
```

test/fft_utest.cpp
```diff
@@ -483,3 +549,244 @@ TYPED_TEST(FftTest, TestFftGf2n) // NOLINT
         this->test_fft_codec(gf, &fft, len);
     }
 }
+
+TYPED_TEST(FftTest, TestFftLarge) // NOLINT
+{
+    unsigned n = 256;
```
slaperche-scality: those can be `const`

test/fft_utest.cpp
```diff
@@ -483,3 +549,244 @@ TYPED_TEST(FftTest, TestFftGf2n) // NOLINT
...
+    std::unique_ptr<gf::Field<TypeParam>> gf = nullptr;
+    std::unique_ptr<quadiron::vec::Vector<TypeParam>> a = nullptr;
+    std::unique_ptr<quadiron::vec::Vector<TypeParam>> _a1 = nullptr;
+    std::unique_ptr<quadiron::vec::Vector<TypeParam>> _a2 = nullptr;
+    std::unique_ptr<fft::FourierTransform<TypeParam>> fft = nullptr;
+    std::unique_ptr<fft::FourierTransform<TypeParam>> fft2 = nullptr;
+
+    gf = gf::alloc<gf::Field<TypeParam>, gf::Prime<TypeParam>>(q);
+
+    // TypeParam p = gf->get_primitive_root();
```
slaperche-scality: dead code, to be removed (same comment for the others)

test/fft_utest.cpp
```diff
@@ -483,3 +549,244 @@ TYPED_TEST(FftTest, TestFftGf2n) // NOLINT
...
+
+    ASSERT_EQ(*_a1 == *_a2, true);
+}
+
+TYPED_TEST(FftTest, TestFftSpeed) // NOLINT
+{
+    std::vector<uint64_t> qs({7681, 12289, 65537});
+    std::vector<uint64_t> ns({256, 512, 1024});
+
+    for (std::vector<uint64_t>::iterator it = qs.begin(); it != qs.end();
```
slaperche-scality: use for-range loop, `for (auto q: sq) {`

test/fft_utest.cpp
```diff
@@ -483,3 +549,244 @@ TYPED_TEST(FftTest, TestFftGf2n) // NOLINT
...
+         it++) {
+        TypeParam q = *it;
+
+        for (std::vector<uint64_t>::iterator it2 = ns.begin(); it2 != ns.end();
```
slaperche-scality: ditto, for-range is more legible

test/fft_utest.cpp
```diff
@@ -483,3 +549,244 @@ TYPED_TEST(FftTest, TestFftGf2n) // NOLINT
...
+             it2++) {
+            TypeParam n = *it2;
+
+            ASSERT_EQ(quadiron::arith::is_prime(q), true);
+
+            std::cout << "choosing q=" << q << " n=" << n << "\n";
```
slaperche-scality: to be removed

test/fft_utest.cpp
```diff
@@ -483,3 +549,244 @@ TYPED_TEST(FftTest, TestFftGf2n) // NOLINT
...
+
+            // TypeParam p = gf->get_primitive_root();
+            // std::cout << "p=" << p << "\n";
+            TypeParam r = gf->get_nth_root(n);
+            // std::cout << "r=" << r << "\n";
+            int l = arith::log2<TypeParam>(n);
+            fft = std::make_unique<fft::Naive<TypeParam>>(*gf, n, r);
+            try {
+                fft2 = std::make_unique<fft::Large<TypeParam>>(*gf, l, r);
+            } catch (...) {
```
slaperche-scality: catch-all is a a bad practice, you should specify what you catch (`quadiron::Exception` for instance)

test/fft_utest.cpp
```diff
@@ -483,3 +549,244 @@ TYPED_TEST(FftTest, TestFftGf2n) // NOLINT
...
+            uint64_t end3 = quadiron::hw_timer();
+
+            if (fft2) {
+                ASSERT_EQ(*_a1 == *_a2, true);
+            }
+            if (fft3) {
+                ASSERT_EQ(*_a1 == *_a3, true);
+            }
+
+            std::cout << "naive: " << end1 - start1;
```
slaperche-scality: removes these `std::cout`: test are not supposed to produce output (and I would even say that unit test are not a good place for benchmarking)

test/fft_utest.cpp
```diff
@@ -483,3 +549,244 @@ TYPED_TEST(FftTest, TestFftGf2n) // NOLINT
...
+                fft3 = nullptr;
+            }
+
+            a = std::make_unique<quadiron::vec::Vector<TypeParam>>(*gf, n);
+            _a1 = std::make_unique<quadiron::vec::Vector<TypeParam>>(*gf, n);
+            _a2 = std::make_unique<quadiron::vec::Vector<TypeParam>>(*gf, n);
+            _a3 = std::make_unique<quadiron::vec::Vector<TypeParam>>(*gf, n);
+            a->rand();
+
+            uint64_t start1 = quadiron::hw_timer();
```
slaperche-scality: `start*` and `end*` variable should be `const`.

Also having n1, n2, n3, … is sign a bad naming and code duplication: maybe we could have a `timeit(func_to_time)` helper that run the specific function and return its execution time.

test/fft_utest.cpp
```diff
@@ -483,3 +549,244 @@ TYPED_TEST(FftTest, TestFftGf2n) // NOLINT
...
+}
+
+TYPED_TEST(FftTest, TestFftLarge2) // NOLINT
+{
+    // Schönhage-Strassen algorithm
+    // Example taken from Pierre Meunier's book
+    if (sizeof(TypeParam) < 8)
+        return;
+
+    int b = 10; // base
```
slaperche-scality: those can be `const`

test/fft_utest.cpp
```diff
@@ -483,3 +549,244 @@ TYPED_TEST(FftTest, TestFftGf2n) // NOLINT
...
+    uint64_t w = arith::chinese_remainder<uint64_t>(2, _a, _n);
+    // std::cerr << " w=" << w << "\n";
+    ASSERT_EQ(w, 25559439);
+
+    std::unique_ptr<gf::Field<TypeParam>> gf_m =
+        gf::alloc<gf::Field<TypeParam>, gf::Prime<TypeParam>>(m);
+    fft::Large<TypeParam> fft(*gf_m, l, 25559439);
+
+    // parse the big numbers
+    char X[] = "1236548787985654354598651354984132468";
```
slaperche-scality: use `str::string`

test/fft_utest.cpp
lamphamsy: +1

src/vec_vector.h
```diff
@@ -459,6 +460,18 @@ void Vector<T>::swap(unsigned i, unsigned j)
     swap(mem[i], mem[j]);
 }
 
+template <typename T>
+void Vector<T>::rand()
+{
+    // XXX this actually does not work because gf_ring->rand()
```
lamphamsy: I try to do it here: #278

src/simd/simd.h
vrancurel: When compiling quadiron-tools:
```
/usr/local/include/quadiron/simd/simd.h:34:10: fatal error: simd/allocator.h: No such file or directory
 #include "simd/allocator.h"
          ^~~~~~~~~~~~~~~~~~
```
I put it aside for now, we need to solve it later.

src/vec_vector.h
vrancurel: this is the current name defined in vec_poly.h

vrancurel: removed the comment

src/vec_vector.h
vrancurel: removed the comment

src/vec_vector.h
slaperche-scality: ho, indeed >_<

test/fft_utest.cpp
```diff
@@ -483,3 +548,204 @@ TYPED_TEST(FftTest, TestFftGf2n) // NOLINT
...
+}
+
+TYPED_TEST(FftTest, TestFftLarge2) // NOLINT
+{
+    // Schönhage-Strassen algorithm
+    // Example taken from Pierre Meunier's book
+    if (sizeof(TypeParam) < 8)
+        return;
+
+    int b = 10; // base
```
NicolasT: If you need a comment explaining `b` is 'base', consider calling the variable `base` instead of `b`.

src/simd/simd.h
vrancurel: actually checked, it is good to do that

test/fft_utest.cpp
```diff
@@ -140,6 +143,67 @@ class FftTest : public ::testing::Test {
...
+     */
+    std::string big_num_add(std::string& str1, std::string& str2)
+    {
+        if (str1.length() > str2.length()) {
+            std::swap(str1, str2);
+        }
+
+        std::string str;
+
+        int n1 = str1.length();
```
slaperche-scality: Return type of `std::string.length` is `size_t` not `int` (I think I mentioned it in a previous review round)

test/fft_utest.cpp
slaperche-scality: What about this comment?

It's weird to have a `add` operation that modifies its input, no?

test/fft_utest.cpp
```diff
@@ -140,6 +143,67 @@ class FftTest : public ::testing::Test {
...
+    /** Convert a number into a vector of digits padded with zeros
+     *
+     * @param vec destination vector
+     * @param n vector size
+     * @param num number in string representation
+     */
+    void big_num_to_vec(vec::Vector<T>& vec, int n, std::string num)
+    {
+        int i;
+        int len = num.length();
```
slaperche-scality: `size_t`, not `int`. And it can be `const`.

test/fft_utest.cpp
```diff
@@ -140,6 +143,67 @@ class FftTest : public ::testing::Test {
...
+        for (i = 0; i < len; i++) {
+            vec.set(i, num[len - i - 1] - '0');
+        }
+        for (; i < n; i++) {
```
slaperche-scality: Could use `std::fill(vec.begin() + i, vec.end(), 0); ` no?

Or `std::fill_n(vec.begin() + i, n - i, 0);` if `n` is not the length of the vector.

test/fft_utest.cpp
slaperche-scality: What about this comment? And the one l.587?

test/fft_utest.cpp
```diff
@@ -483,3 +544,211 @@ TYPED_TEST(FftTest, TestFftGf2n) // NOLINT
...
+}
+
+TYPED_TEST(FftTest, TestFftLarge2) // NOLINT
+{
+    // Schönhage-Strassen algorithm
+    // Example taken from Pierre Meunier's book
+    if (sizeof(TypeParam) < 8)
+        return;
+
+    int base = 10;
```
slaperche-scality: Those can be `const` no?

test/fft_utest.cpp
vrancurel: pass now by value. this is just a test and a straightforward addition. Don't want to optimize here.

test/fft_utest.cpp
vrancurel: it is a special operation where we embed the string into the vector, we must set the ascii zeroes and pad by zeroes.

test/fft_utest.cpp
vrancurel: see comment below

test/fft_utest.cpp
vrancurel: since we cannot transform the above loop this is weird.

slaperche-scality: > since we cannot transform the above loop this is weird.
Why?

I don't see why keeping the above loop as-is prevent to use more explicit function when it's possible.

test/fft_utest.cpp
slaperche-scality: Ok, since it's only for test it's fine by me

test/fft_utest.cpp
lamphamsy: There's not `vec.begin()`, `vec.end` as `vec` is `quadiron::vec::Vector`, not `std::vector`. We could use `std::fill` or `std::fill_n` but we need to get the array handled by `vec`. 
So I think we can keep it like this.

slaperche-scality: Ha since it's our own array with it weird interface, yeah we can keep it as is…

PR #278, opened 2019-02-21
GF: init random generator in constructor

lamphamsy:
Random number generator is initialised in the constructor, and its seed is a random number

### DIFF
--- a/src/gf_ring.h
+++ b/src/gf_ring.h
@@ -148,12 +148,30 @@ class RingModN {
     std::vector<int> exponents;
     std::vector<T> all_primes_factors;
     std::vector<T> proper_divisors;
+    // random number engine with a seed as a random number
+    mutable std::uniform_int_distribution<uint64_t> distribution;
+    // random number engine for the __uint128_t case
+    mutable std::uniform_int_distribution<uint64_t> distribution_leading;
+
+  private:
+    static constexpr uint64_t rand_upper_bound(T card)
+    {
+        return (card > std::numeric_limits<uint64_t>::max())
+                   ? std::numeric_limits<uint64_t>::max()
+                   : card - 1;
+    }
+    static constexpr uint64_t rand_leading_bound(T card)
+    {
+        return (card > std::numeric_limits<uint64_t>::max())
+                   ? card - std::numeric_limits<uint64_t>::max() - 1
+                   : 0;
+    }
 };
 template <typename T>
 RingModN<T>::RingModN(T card)
+    : _card(card), root(0), distribution(1, rand_upper_bound(card)),
+      distribution_leading(0, rand_leading_bound(card))
 {
-    this->_card = card;
-    this->root = 0;
 }
 
 template <typename T>
@@ -520,8 +538,17 @@ void RingModN<T>::compute_omegas_cached(vec::Vector<T>& W, int n, T w) const
 template <typename T>
 T RingModN<T>::rand(void) const
 {
-    std::uniform_int_distribution<uint32_t> dis(1, this->card() - 1);
-    return dis(prng());
+    return distribution(prng());
+}
+
+template <>
+inline __uint128_t RingModN<__uint128_t>::rand(void) const
+{
+    __uint128_t c = distribution(prng());
+    if (_card > std::numeric_limits<uint64_t>::max()) {
+        c |= (static_cast<__uint128_t>(distribution_leading(prng())) << 64);
+    }
+    return c;
 }
 
 /** Check if a number is primitive root or not.
--- a/test/gf_utest.cpp
+++ b/test/gf_utest.cpp
@@ -116,6 +116,38 @@ class GfTestCommon : public ::testing::Test {
 using AllTypes = ::testing::Types<uint32_t, uint64_t, __uint128_t>;
 TYPED_TEST_CASE(GfTestCommon, AllTypes);
 
+TYPED_TEST(GfTestCommon, TestGfRing) // NOLINT
+{
+    std::vector<TypeParam> cards = {
+        257,        // Fermat 2^8 + 1
+        65537,      // Fermat 2^16 + 1
+        2147483647, // Mersenne 2^31 - 1
+    };
+
+    if (sizeof(TypeParam) >= sizeof(uint64_t)) {
+        // 9th Mersenne 2^61 - 1
+        cards.push_back((static_cast<TypeParam>(1) << 61) - 1);
+    } else if (sizeof(TypeParam) >= sizeof(__uint128_t)) {
+        // Mersenne 2^89 - 1
+        cards.push_back((static_cast<TypeParam>(1) << 89) - 1);
+    }
+
+    for (auto& card : cards) {
+        auto gf(gf::create<gf::Field<TypeParam>>(card, 1));
+        this->test_negation(gf);
+        this->test_find_primitive_root(&gf);
+        // NOTE: Following tests does not work for __uint128_t as in `mul` it
+        // use DoubleSizeVal<TypeParam> this->test_log(gf);
+        if (sizeof(TypeParam) < 16) {
+            this->test_reciprocal(gf);
+            this->test_get_order(gf);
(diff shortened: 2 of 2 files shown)


src/gf_ring.h
```diff
@@ -148,12 +148,13 @@ class RingModN {
     std::vector<int> exponents;
     std::vector<T> all_primes_factors;
     std::vector<T> proper_divisors;
+    // random number engine with a seed as a random number
+    mutable std::default_random_engine generator{std::random_device()()};
+    mutable std::uniform_int_distribution<T> distribution;
```
slaperche-scality: You can't use `T`, 128-bit is not standard and thus not supported.

Also why mutable? And the init should probably be done in the constructor initialization list, not here.

lamphamsy: > You can't use `T`, 128-bit is not standard and thus not supported.

Yep, we can't use `T` here :(

> Also why mutable? 

I set mutable as `distribution` and `generator` will be called from a const instance of GF, while their state will be updated after each call.

> And the init should probably be done in the constructor initialization list, not here.

Yes, they will be updated :)

vrancurel: can you use PRNG() ?

lamphamsy: it's done

test/gf_utest.cpp
```diff
@@ -116,6 +116,45 @@ class GfTestCommon : public ::testing::Test {
 using AllTypes = ::testing::Types<uint32_t, uint64_t, __uint128_t>;
 TYPED_TEST_CASE(GfTestCommon, AllTypes);
 
+TYPED_TEST(GfTestCommon, TestGfRing) // NOLINT
+{
+    const TypeParam mersenne_9 =
+        ((((static_cast<TypeParam>(1) << 16) << 16) << 16) << 13)
```
slaperche-scality: Why not `<< 61` directly? BTW, how does that work when `TypeParam` is 32-bit?

You'll have an undefined behavior (see section 5.8 of the C++ standard):

> The behavior is undefined if the right operand is negative, or greater than or equal to the length in bits of the promoted left operand.

(maybe you avoid the UB, by doing little shift but it's obfuscating the code)
Ditto for `mersenne_10`.

I think you could rewrite the code as

```
    TypeParam cards[5] = {
        257,        // Fermat 2^8 + 1
        65537,      // Fermat 2^16 + 1
        2147483647, // Mersenne 2^31 - 1
        0,
        0
    };

    int cards_nb_test = 3;
    if (sizeof(TypeParam) >= sizeof(uint64_t)) {
        // 9th Mersenne 2^61 - 1
        cards[cards_nb_test++] = (static_cast<TypeParam>(1) << 61) - 1;
        cards_nb_test++;
    } else if (sizeof(TypeParam) >= sizeof(__uint128_t)) {
        // Mersenne 2^89 - 1
        cards[cards_nb_test++] = (static_cast<TypeParam>(1) << 89) - 1;
    }
```

It would be more readable.

src/gf_ring.h
```diff
@@ -148,12 +148,24 @@ class RingModN {
     std::vector<int> exponents;
     std::vector<T> all_primes_factors;
     std::vector<T> proper_divisors;
+    // random number engine with a seed as a random number
+    mutable std::uniform_int_distribution<uint64_t> distribution;
+    // random number engine for the __uint128_t case
+    mutable std::uniform_int_distribution<uint64_t> distribution_leading;
```
slaperche-scality: Why do you need this one? You could simply make two calls to `distribution` for the 128-bit case, no?

src/gf_ring.h
```diff
@@ -520,8 +532,17 @@ void RingModN<T>::compute_omegas_cached(vec::Vector<T>& W, int n, T w) const
...
-    return dis(prng());
+    return distribution(prng());
+}
+
+template <>
+inline __uint128_t RingModN<__uint128_t>::rand(void) const
+{
+    __uint128_t c = distribution(prng());
+    if (_card > std::numeric_limits<uint64_t>::max()) {
+        c |= ((((distribution_leading(prng()) << 16) << 16) << 16) << 16);
```
slaperche-scality: could be simplified as `c |= (static_cast<__uint128_t>(distribution(prng())) << 64);`

src/gf_ring.h
```diff
@@ -148,12 +148,24 @@ class RingModN {
...
+    // random number engine with a seed as a random number
+    mutable std::uniform_int_distribution<uint64_t> distribution;
+    // random number engine for the __uint128_t case
+    mutable std::uniform_int_distribution<uint64_t> distribution_leading;
 };
 template <typename T>
 RingModN<T>::RingModN(T card)
+    : _card(card), root(0), distribution(
+                                1,
+                                (card > std::numeric_limits<uint64_t>::max())
```
slaperche-scality: This ternary could be moved into a little helper to improve the readability:

```
[…]
  private:
    static constexpr uint64_t rand_upper_bound(T card)
    {
        return (card > std::numeric_limits<uint64_t>::max())
                   ? std::numeric_limits<uint64_t>::max()
                   : card - 1;
    }
};
template <typename T>
RingModN<T>::RingModN(T card)
    : _card(card), root(0), distribution(1, rand_upper_bound(card))
{
}

[…]
```

test/gf_utest.cpp
slaperche-scality: Or even simpler:

```C++
    std::vector<TypeParam> cards = {
        257,        // Fermat 2^8 + 1
        65537,      // Fermat 2^16 + 1
        2147483647, // Mersenne 2^31 - 1
    };

    if (sizeof(TypeParam) >= sizeof(uint64_t)) {
        // 9th Mersenne 2^61 - 1
        cards.push_back((static_cast<TypeParam>(1) << 61) - 1);
    } else if (sizeof(TypeParam) >= sizeof(__uint128_t)) {
        // Mersenne 2^89 - 1
        cards.push_back((static_cast<TypeParam>(1) << 89) - 1);
    }

    for (auto& card : cards) {
        auto gf(gf::create<gf::Field<TypeParam>>(card, 1));
        […]
    }
```

lamphamsy: yes, it's better. 
However, we cannot use `<< 61` or `<<89` as it will raise an error. We avoided it somewhere else with the same reason.

src/gf_ring.h
lamphamsy: The two distributions are used for two different purposes.

- If `q <= std::numeric_limits<uint64_t>::max()`, only `distribution` is enough.

- Otherwise,  we use `distribution` and `distribution_leading` to generate random number `x = x1 + x2` where

    - `x1 in [1, std::numeric_limits<uint64_t>::max()` is generated by `distribution`
    - `x2 in [0, q-std::numeric_limits<uint64_t>::max()`) is generated by`distribution_leading`
 
Oups, I saw a bug to fix :)

test/gf_utest.cpp
slaperche-scality: > However, we cannot use << 61 or <<89 as it will raise an error.

Are you sure?
IIRC, I tried the code I proposed here and it worked fine.

src/gf_ring.h
slaperche-scality: Ha, indeed, then it makes sense to have two distribution.

test/gf_utest.cpp
lamphamsy: oh, you're right. It works. Thanks.
My previous check fails as I use `1` instead of `static_cast<TypeParam>(1)`.

src/gf_ring.h
```diff
@@ -148,12 +148,24 @@ class RingModN {
...
+                                    ? std::numeric_limits<uint64_t>::max()
+                                    : card - 1),
+      distribution_leading(
+          1,
```
lamphamsy: - [ ] fix starting from 0

src/gf_ring.h
lamphamsy: yeah, thank you

PR #279, opened 2019-03-26
Fix bugs in fft::Additive

lamphamsy:
- Fix fft::Additive: cover more cases of FFT length
    - Cover all cases of `t >= 2` and `n >= 1` for Taylor expansions and inverse ones
    - Add more cases of `n`, `t`, `code length` in testing

- Fft::Additive uses smart pointers
- Fft::Additive: do always modify input after Taylor expansion
- Fix vec::Vector's copy constructor

### DIFF
--- a/src/fft_add.h
+++ b/src/fft_add.h
@@ -53,8 +53,10 @@ class Additive : public FourierTransform<T> {
     using FourierTransform<T>::ifft;
     using FourierTransform<T>::fft_inv;
 
-    Additive(const gf::Field<T>& gf, T m, vec::Vector<T>* betas = nullptr);
-    ~Additive();
+    Additive(
+        const gf::Field<T>& gf,
+        T m,
+        std::shared_ptr<vec::Vector<T>> betas = nullptr);
     void compute_basis();
     void compute_beta_m_powers();
     void compute_G();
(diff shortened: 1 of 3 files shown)


src/fft_add.h
```diff
@@ -83,21 +85,24 @@ class Additive : public FourierTransform<T> {
...
-    vec::Vector<T>* deltas = nullptr;
-    vec::Vector<T>* beta_m_powers = nullptr;
-    vec::Vector<T>* G = nullptr;
-    vec::Vector<T>* g0 = nullptr;
-    vec::Vector<T>* g1 = nullptr;
-    vec::Vector<T>* u = nullptr;
-    vec::Vector<T>* v = nullptr;
-    vec::Vector<T>* mem = nullptr;
-    Additive<T>* fft_add = nullptr;
+    std::shared_ptr<vec::Vector<T>> betas = nullptr;
```
slaperche-scality: Do we really share the ownership of those vector?

Since they are created by this class, I would expect it to keep the exclusive ownership of the memory (and lending reference to functions that need to use them, but noy sharing the ownership).

src/fft_add.h
```diff
@@ -122,55 +127,28 @@ Additive<T>::Additive(const gf::Field<T>& gf, T m, vec::Vector<T>* betas)
...
 
-        this->u = new vec::Vector<T>(gf, this->m_k);
-        this->v = new vec::Vector<T>(gf, this->m_k);
+        this->u = std::make_unique<vec::Vector<T>>(gf, this->m_k);
+        this->v = std::make_unique<vec::Vector<T>>(gf, this->m_k);
 
-        this->mem = new vec::Vector<T>(gf, this->n);
-        this->fft_add = new Additive(gf, m - 1, this->deltas);
+        this->mem = std::make_unique<vec::Vector<T>>(gf, this->n);
+        this->fft_add = std::make_unique<Additive>(gf, m - 1, this->deltas);
```
slaperche-scality: Is that why you need `shared_ptr` for `deltas`? I think you could pass a reference here because since the `fft_add` is a member of this class, it is guaranteed to have a shorter lifetime than `this`, so we are sure that it won't outlive `this` (and thus `deltas`), no?

lamphamsy: `deltas` will be used as `betas` in `this->fft_add`. And we want to use a smart pointer for `betas`. Is there a better solution for `betas` and `deltas`?

slaperche-scality: OK for `shared_pointer` since it's a kind of recursive classes and I don't have any better idea right now.

lamphamsy: yes, it is. Thanks 👍

PR #280, opened 2019-04-09
Build Release mode with enabled SIMD by default

lamphamsy:
Observing a timeout test in liberasurecodes'  unit test, we could avoid such cases by using Release + SIMD mode by default

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,7 +50,7 @@ jobs:
           name: Compile on ARM64 with the NDK
           command: >
             rm -rf build && mkdir build && cd build &&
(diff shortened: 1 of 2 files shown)


slaperche-scality:
> Observing a timeout test in liberasurecodes' unit test, we could avoid such cases by using Release + SIMD mode by default

Can't we configure how the lib is compiled for the test? Instead of changing the defaults.

lamphamsy:
@slaperche-scality : yes we can configure the lib by specifying flags. But the purpose of this change is to make quadiron simple to use :)

slaperche-scality:
OK, fine by me

PR #281, opened 2019-04-11
Rearrange SIMD parts

lamphamsy:
- Rename `simd_fnt.h` to `simd_radix2_fft.h` includes operations dedicated for radix-2 FFT
- Move accelerated operations dedicated for RingModN from `simd_basic.h` to a new file `simd_ring.h`
- Rename `simd_basic.h` to `simd_fnt.h` includes operations dedicated for Fermat GF(q)


src/simd_basic_fnt.h
```diff
@@ -28,8 +28,8 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
```
slaperche-scality: Why `simd_basic_fnt` instead of simply `simd_fnt`?

src/simd_basic_fnt.h
```diff
@@ -198,135 +198,6 @@ inline void add_props(
     }
 }
 
-/* ==================== Operations for RingModN =================== */
```
slaperche-scality: This commit could be merged with ` [New] simd_ring.h includes accelerated operations dedicated for RingModN ` no?
You are onyl moving code, not modifying it right?

If yes, it's clearer to do the move in a single commit.

src/simd.h
```diff
@@ -60,10 +60,13 @@ namespace simd {
 #endif
 
 // Include basic operations
-#include "simd_basic.h"
+#include "simd_basic_fnt.h"
```
slaperche-scality: Comment above should be updated to `Include FNT operations.` or `Include accelerated operations for FNT.`

src/simd_basic_fnt.h
lamphamsy: yes, okay. I group commits concerning splitting `simd_basic` to `simd_fnt` and `simd_ring` into one commit.

PR #283, opened 2019-05-09
Buffers stores data and meta

lamphamsy:
`Buffers` consists of a vector of `n` DATA buffers (array of T) and an OPTIONAL vector of `n` META buffers (array of `uint8_t`). A meta buffer corresponds to a data buffer. Every `s`-byte data element of a data buffer corresponds to an `s`-bit element of the meta buffer. A pair of data and meta elements can
represent an integer of `8*s + s`-bits Each data element points to a buffer of size `m * sizeof(T)` bytes.
Hence, each meta element points to a buffer of size `bsize = m * sizeof(T) / 8` bytes that should be an integer, i.e. `m * sizeof(T) % 8 = 0`. This condition is not difficult to achieve as a convenient `size` can be always chosen with a negligible impact on its application.

### DIFF
--- a/src/vec_buffers.h
+++ b/src/vec_buffers.h
@@ -153,9 +153,88 @@ class Buffers final {
         unsigned data_len,
         unsigned group_len);
     void radix2_fft_inv_prepare(const Buffers<T>& input);
+    void reset_meta();
     friend bool operator==<T>(const Buffers<T>& lhs, const Buffers<T>& rhs);
     void dump(void);
     void swap(unsigned i, unsigned j);
+
+    /** Calculate meta size given a buffer size
+     *  meta size := ceil(size * sizeof(T) / CHAR_BIT)
+     *
+     * @param s - given size, in words
+     * @return meta size, in bytes
+     */
+    static size_t compute_meta_size(size_t s)
+    {
+        assert(s > 0);
+
+        const size_t bytes = s * sizeof(T);
+        size_t r = bytes / CHAR_BIT;
+        while (r * CHAR_BIT < bytes) {
+            r++;
+        }
+        return r;
+    };
+
+    /** Calculate buffer size given a meta_size
+     *  buffer size := ceil(meta_size * CHAR_BIT / sizeof(T))
+     *
+     * @param m_size - given meta size, in bytes
+     * @return meta size
+     */
+    static size_t compute_size(size_t m_size)
+    {
+        assert(m_size > 0);
+
+        const size_t bytes = m_size * CHAR_BIT;
+        size_t r = bytes / sizeof(T);
+        while (r * sizeof(T) < bytes) {
+            r++;
+        }
+        return r;
+    };
+
+    /** Calculate conventional size of buffers
+     * Conentional size is the lowest number of words that is at least a given
+     * `s` and satisfy the following conditions:
+     *  - its bytes is multiple of `size_alignment`
+     *  - the correspondent `meta_size` is multiple of `meta_size_alignment`
+     *
+     * @param s - given size, in words
+     * @param size_alignment - alignment number of output size, in bytes
+     * @param meta_size_alignment - alignment number of meta size according to
+     * the out size, in byte
+     * @return conventional size, in words
+     */
+    static size_t get_conv_size(
+        size_t s,
+        size_t size_alignment = 0,
+        size_t meta_size_alignment = 0)
+    {
+        assert(s > 0);
+
+        if (size_alignment == 0) {
+            size_alignment = simd::ALIGNMENT;
+        }
+
+        if (meta_size_alignment == 0) {
+            // it's the `meta_size` for a single element fitting a Register
+            meta_size_alignment = simd::ALIGNMENT / CHAR_BIT;
+        }
+
+        // calculate meta_size
+        size_t m_size = Buffers<T>::compute_meta_size(s);
+        // calculate new size according to `m_size`
+        size_t c_size = Buffers<T>::compute_size(m_size);
+
+        while (c_size * sizeof(T) % size_alignment != 0
+               || m_size % meta_size_alignment != 0) {
+            m_size++;
+            c_size = Buffers<T>::compute_size(m_size);
+        }
+
+        return c_size;
+    };
 
   protected:
     std::vector<T*> mem;
@@ -173,7 +252,7 @@ class Buffers final {
     uint8_t* zeros_meta = nullptr;
     bool m_meta = false;
 
-    void compute_meta_size();
+    void init_meta();
     void allocate_meta(bool init_zero = false);
 };
 
@@ -198,7 +277,7 @@ Buffers<T>::Buffers(int n, size_t size, bool has_meta)
 
     if (has_meta) {
         this->m_meta = has_meta;
-        this->compute_meta_size();
+        this->init_meta();
         this->allocate_meta(true);
     }
 }
@@ -225,7 +304,7 @@ Buffers<T>::Buffers(
 
     if (meta) {
         this->m_meta = true;
-        this->compute_meta_size();
+        this->init_meta();
         this->meta = *meta;
     }
 }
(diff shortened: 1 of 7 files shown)


src/vec_buffers.h
```diff
@@ -632,6 +639,34 @@ inline const T* Buffers<T>::get(int i) const
...
+ * - low half part is the low (or high) part of the `j`th element
+ * - high half part is the low (or high) part of the meta of the `j`th element
+ *
+ * @param buf_id - index of buffer
+ * @param ele_id - index of element in the ith buffer
+ * @param hi - high half part of the unpacked element
+ * @param lo - low half part of the unpacked element
+ */
+template <typename T>
+void Buffers<T>::get(int buf_id, size_t ele_id, T& hi, T& lo) const
```
slaperche-scality: See my comments on `Buffers<T>::set`, most of them also apply here.

src/vec_buffers.h
```diff
@@ -658,6 +693,47 @@ inline const std::vector<uint8_t*>& Buffers<T>::get_meta() const
...
 }
 
+/** Get meta value of the `j`th element at the `i`th buffer
+ *
+ * @param buf_id - index of buffer
+ * @param ele_id - index of element in the ith buffer
+ * @return meta value
+ */
+template <typename T>
+T Buffers<T>::get_meta(int buf_id, size_t ele_id) const
```
slaperche-scality: Ditto, see my comment on `Buffers<T>::set_meta`, most of them also apply here.

src/vec_buffers.h
```diff
@@ -651,6 +657,69 @@ void Buffers<T>::swap(unsigned i, unsigned j)
     }
 }
 
+/** Perform a preparatiion in radix-2 FFT algorithm
```
slaperche-scality: s/tiion/tion/

src/vec_buffers.h
```diff
@@ -651,6 +657,69 @@ void Buffers<T>::swap(unsigned i, unsigned j)
...
+ *
+ * @param input - input buffers
+ * @param scrambler - a vector stores bit-reversed values
+ * @param data_len - a conven length
```
slaperche-scality: conven?

src/vec_buffers.h
```diff
@@ -651,6 +657,69 @@ void Buffers<T>::swap(unsigned i, unsigned j)
...
+        for (unsigned i = scrambler[idx]; i < scrambler[idx] + group_len; ++i) {
+            std::copy_n(i_mem[idx], size, mem[i]);
+            m_meta ? std::copy_n(i_meta[idx], meta_size, meta[i]) : nullptr;
+        }
+    }
+    for (unsigned idx = input_len; idx < data_len; ++idx) {
+        // set output  = scramble(input), i.e. bit reversal ordering
+        for (unsigned i = scrambler[idx]; i < scrambler[idx] + group_len; ++i) {
+            std::fill_n(mem[i], size, 0);
+            m_meta ? std::fill_n(meta[i], meta_size, 0) : nullptr;
```
slaperche-scality: Ternary are usually used for assignment. That's not the case here.

Could would be clearer written as:

```
if (m_meta) {
    std::fill_n(meta[i], meta_size, 0)
}
```

src/vec_buffers.h
```diff
@@ -651,6 +657,69 @@ void Buffers<T>::swap(unsigned i, unsigned j)
...
+    for (i = 0; i < input_len; ++i) {
+        std::copy_n(i_mem[i], size, mem[i]);
+        m_meta ? std::copy_n(i_meta[i], meta_size, meta[i]) : nullptr;
+    }
+
+    if (input_len < len) {
+        const unsigned input_len_power_2 = arith::ceil2<unsigned>(input_len);
+        for (; i < input_len_power_2; ++i) {
+            std::fill_n(mem[i], size, 0);
+            m_meta ? std::fill_n(meta[i], meta_size, 0) : nullptr;
```
slaperche-scality: same remarks, don't use a ternary here (an `if` is clearer)

test/buffers_utest.cpp
```diff
@@ -362,3 +362,176 @@ TYPED_TEST(BuffersTest, TestCalculateSize) // NOLINT
         ASSERT_EQ(t.expected[id], new_size);
     }
 }
+
+TYPED_TEST(BuffersTest, TestGetValueAndMeta) // NOLINT
```
slaperche-scality: If possible, try to have two test functions: one to test values, and one to test metadata (it will make test more like unit test, since you'll be testing a single function and also the test will be shorter => more readable).

src/vec_buffers.h
```diff
@@ -720,6 +801,17 @@ void Buffers<T>::radix2_fft_inv_prepare(const Buffers<T>& input)
     }
 }
 
+/// Reset meta buffers
+template <typename T>
+inline void Buffers<T>::reset_meta()
+{
+    if (m_meta) {
+        for (int i = 0; i < n; i++) {
+            std::fill_n(meta[i], meta_size, 0);
```
slaperche-scality: Any reason why you are using `0` instead of reusing your `zero_meta` vector?

src/vec_buffers.h
```diff
@@ -153,10 +153,89 @@ class Buffers final {
...
+     * @param s - given size, in words
+     * @return meta size, in bytes
+     */
+    static size_t compute_meta_size(size_t s)
+    {
+        assert(s > 0);
+
+        const size_t bytes = s * sizeof(T);
+        size_t r = bytes / CHAR_BIT;
+        while (r * CHAR_BIT < bytes) {
```
slaperche-scality: see my previous comment on that `while`.

src/vec_buffers.h
```diff
@@ -153,10 +153,89 @@ class Buffers final {
...
+     * @param m_size - given meta size, in bytes
+     * @return meta size
+     */
+    static size_t compute_size(size_t m_size)
+    {
+        assert(m_size > 0);
+
+        const size_t bytes = m_size * CHAR_BIT;
+        size_t r = bytes / sizeof(T);
+        while (r * sizeof(T) < bytes) {
```
slaperche-scality: ditto about the while

src/vec_buffers.h
```diff
@@ -153,10 +153,89 @@ class Buffers final {
...
+        return r;
+    };
+
+    /** Calculate buffer size given a meta_size
+     *  buffer size := ceil(meta_size * CHAR_BIT / sizeof(T))
+     *
+     * @param m_size - given meta size, in bytes
+     * @return meta size
+     */
+    static size_t compute_size(size_t m_size)
```
slaperche-scality: `compute_size` and `compute_meta_size` are too similar, you should have a single function instead of copy/pasting.

src/vec_buffers.h
```diff
@@ -153,10 +153,89 @@ class Buffers final {
...
+        const size_t bytes = m_size * CHAR_BIT;
+        size_t r = bytes / sizeof(T);
+        while (r * sizeof(T) < bytes) {
+            r++;
+        }
+        return r;
+    };
+
+    /** Calculate conventional size of buffers
+     * Conentional size is the lowest number of words that is at least a given
```
slaperche-scality: s/Conentional/conventional/

Also, skip a line between the short description (the first line) and the detailed one (the rest).

src/vec_buffers.h
```diff
@@ -153,10 +153,89 @@ class Buffers final {
...
+     * `s` and satisfy the following conditions:
+     *  - its bytes is multiple of `size_alignment`
+     *  - the correspondent `meta_size` is multiple of `meta_size_alignment`
+     *
+     * @param s - given size, in words
+     * @param size_alignment - alignment number of output size, in bytes
+     * @param meta_size_alignment - alignment number of meta size according to
+     * the out size, in byte
+     * @return conventional size, in words
```
slaperche-scality: What's the size of word? is it 32-bit or is it `sizeof(T)`?

PR #282, opened 2019-04-25
Fec FNT: enhancement by focusing on `Buffers`

lamphamsy:

PR #284, opened 2019-06-12
Simplify SIMD operations

lamphamsy:
NB: the PR is split/extracted from the big one #282 

The purpose is to eliminate as much as possible branches, particularly in essential operations s.t. modular operations.

Currently, we support vectorized operations of FNT in the cases:
- FNT(257) using `uint16_t` and `uint32_t`
- FNT(65537) using `uint32_t`

But the support of FNT(257) using `uint32_t` raises additional branches of checking the cardinal of the field. And its encoding/decoding speeds are obviously smaller than the two other cases: FNT(257) using `uin16_t` and FNT(65537) using `uint32_t`.

Hence, for FNT, we support only  FNT(257) using `uin16_t` and FNT(65537) using `uint32_t`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -191,7 +191,7 @@ endif()
 
 # Release flags
 set(RELEASE_CXX_FLAGS
-  -O3 -DNDEBUG
+  -O3 -DNDEBUG -flto=full
 )
 foreach(flag ${RELEASE_CXX_FLAGS})
   check_cxx_compiler_flag(${flag} has_flag_${flag})
--- a/scripts/test_ec.sh
+++ b/scripts/test_ec.sh
@@ -174,7 +174,6 @@ do
     fec_type=$(echo $i|cut -d_ -f1)
     word_size=$(echo $i|cut -d_ -f2)
 
-    do_test enconly ${fec_type} ${word_size} 50 50 "" ""
     do_test all ${fec_type} ${word_size} 3 3 "" ""
     do_test all ${fec_type} ${word_size} 3 3 "0 1" "0"
     do_test all ${fec_type} ${word_size} 3 5 "0 1" "0"
--- a/src/simd_128.h
+++ b/src/simd_128.h
@@ -40,29 +40,23 @@ typedef __m128i VecType;
 
 /* ============= Constant variable  ============ */
 
-// @note: using const leads to an lint error of initialization of 'variable'
-// with static storage duration may throw an exception that cannot be caught
+template <typename T>
+inline VecType one();
+template <>
+inline VecType one<uint16_t>()
+{
+    return _mm_set1_epi16(1);
+}
+template <>
+inline VecType one<uint32_t>()
+{
+    return _mm_set1_epi32(1);
+}
 
-// NOLINTNEXTLINE(cert-err58-cpp)
-const VecType F4_U32 = _mm_set1_epi32(65537);
-// NOLINTNEXTLINE(cert-err58-cpp)
-const VecType F4_MINUS_ONE_U32 = _mm_set1_epi32(65536);
-// NOLINTNEXTLINE(cert-err58-cpp)
-const VecType F3_U32 = _mm_set1_epi32(257);
-// NOLINTNEXTLINE(cert-err58-cpp)
-const VecType F3_MINUS_ONE_U32 = _mm_set1_epi32(256);
-
-// NOLINTNEXTLINE(cert-err58-cpp)
-const VecType F3_U16 = _mm_set1_epi16(257);
-// NOLINTNEXTLINE(cert-err58-cpp)
-const VecType F3_MINUS_ONE_U16 = _mm_set1_epi16(256);
-
-// NOLINTNEXTLINE(cert-err58-cpp)
-const VecType ZERO = _mm_setzero_si128();
-// NOLINTNEXTLINE(cert-err58-cpp)
-const VecType ONE_U16 = _mm_set1_epi16(1);
-// NOLINTNEXTLINE(cert-err58-cpp)
-const VecType ONE_U32 = _mm_set1_epi32(1);
+inline VecType zero()
+{
+    return _mm_setzero_si128();
+}
 
 // NOLINTNEXTLINE(cert-err58-cpp)
 const VecType MASK8_LO = _mm_set1_epi16(0x80);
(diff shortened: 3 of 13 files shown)


src/simd_256.h
```diff
@@ -56,26 +56,23 @@ typedef __m128i HalfVecType;
 // @note: using const leads to an lint error of initialization of 'variable'
 // with static storage duration may throw an exception that cannot be caught
```
slaperche-scality: ditto, this comment can be removed I think.

src/simd_fnt.h
```diff
@@ -37,42 +37,55 @@ namespace quadiron {
...
+template <>
+inline VecType get_low_half<uint16_t>(const VecType& x)
+{
+    return BLEND8(zero(), x, MASK8_LO);
+}
+template <>
+inline VecType get_low_half<uint32_t>(const VecType& x)
 {
-    return (q == F3) ? BLEND8(ZERO, x, MASK8_LO) : BLEND16(ZERO, x, 0x55);
+    return BLEND16(zero(), x, 0x55);
```
slaperche-scality: Could we have a document constant (like we did with `MASK8_LO`) instead of having a raw magic number?

src/simd_fnt.h
```diff
@@ -37,42 +37,55 @@ namespace quadiron {
...
+inline VecType get_high_half<uint16_t>(const VecType& x)
+{
+    return BLEND8(zero(), SHIFTR(x, 1), MASK8_LO);
+}
+template <>
+inline VecType get_high_half<uint32_t>(const VecType& x)
 {
-    return (q == F3) ? BLEND8(ZERO, SHIFTR(x, 1), MASK8_LO)
-                     : BLEND16(ZERO, SHIFTR(x, 2), 0x55);
+    return BLEND16(zero(), SHIFTR(x, 2), 0x55);
```
slaperche-scality: ditto (named constant instead of 0x55)

src/simd_radix2_fft.h
```diff
@@ -36,6 +36,24 @@
...
 namespace simd {
 
+enum class CtGsCase {
+    SIMPLE,
+    NORMAL,
+    EXTREME,
+};
+
+template <typename T>
+inline CtGsCase get_case(T r, T q)
```
slaperche-scality: This function can be declared `constexpr`.

e.g.:
```
template <typename T>
constexpr CtGsCase get_case(T r, T q)
```

(`constexpr` implies `inline` so no need to keep it).

src/simd_radix2_fft.h
```diff
@@ -44,22 +62,31 @@ namespace simd {
...
+    case CtGsCase::SIMPLE:
+        y = mod_sub<T>(x, z);
+        x = mod_add<T>(x, z);
+        break;
+    case CtGsCase::EXTREME:
+        y = mod_add<T>(x, z);
+        x = mod_sub<T>(x, z);
+        break;
+    case CtGsCase::NORMAL:
+    default:
```
slaperche-scality: Remove the `default` case.

Since we are using an enum, the compiler can statically enforce that we cover all the enum values.
If one day we add a new case, we will have a compiler error if we don't update the switch, e.g.:

```
error: enumeration value 'NEW_CASE' not handled in switch [-Werror=switch-enum]
```

src/simd_radix2_fft.h
```diff
@@ -69,25 +96,30 @@ inline void butterfly_ct(T rp1, VecType c, VecType* x, VecType* y, T q)
...
+    VecType add = mod_add<T>(x, y);
+    switch (gs_case) {
+    case CtGsCase::SIMPLE:
+        y = mod_sub<T>(x, y);
+        break;
+    case CtGsCase::EXTREME:
+        y = mod_sub<T>(y, x);
+        break;
+    case CtGsCase::NORMAL:
+    default:
```
slaperche-scality: ditto: remove `default`

src/simd_radix2_fft.h
```diff
@@ -96,24 +128,25 @@ inline void butterfly_gs(T rp1, VecType c, VecType* x, VecType* y, T q)
...
-        return mod_neg(x, q);
+    switch (gs_case) {
+    case CtGsCase::EXTREME:
+        x = mod_neg<T>(x);
+        break;
+    case CtGsCase::NORMAL:
+        x = mod_mul<T>(c, x);
+        break;
+    case CtGsCase::SIMPLE:
+    default:
```
slaperche-scality: ditto: remove `default`

test/quadiron_c_utest.cpp
```diff
@@ -213,6 +197,39 @@ class QuadironCTest : public ::testing::Test {
...
+            }
+        }
+
+        for (int i = 0; i < n_parities; i++) {
+            if (missing_idxs[n_data + i]) {
+                std::fill_n(_parity[i], block_size + metadata_size, 0);
+            }
+        }
+
+        // FIXME: for non-systematic FNT, `quadiron_fnt32_decode` will
```
slaperche-scality: IMHO, it's cleaner to have a copy of `_data`/call `quadiron_fnt32_decode` on a copy instead of relying on the order of the test.

CMakeLists.txt
```diff
@@ -191,7 +191,7 @@ endif()
 
 # Release flags
 set(RELEASE_CXX_FLAGS
-  -O3 -DNDEBUG
```
slaperche-scality: So now it works? Great!

We probably have fixed some issue in our code.

scripts/test_ec.sh
```diff
@@ -174,7 +174,7 @@ do
     fec_type=$(echo $i|cut -d_ -f1)
     word_size=$(echo $i|cut -d_ -f2)
 
-    do_test enconly ${fec_type} ${word_size} 50 50 "" ""
+#    do_test enconly ${fec_type} ${word_size} 50 50 "" ""
```
slaperche-scality: Why not removing it totally (it will not be lost, thanks to Git history), and explain why we remove it in the commit message.

src/simd_128.h
```diff
@@ -43,26 +43,23 @@ typedef __m128i VecType;
 // @note: using const leads to an lint error of initialization of 'variable'
 // with static storage duration may throw an exception that cannot be caught
```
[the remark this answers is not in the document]
lamphamsy: Yes, it should be. Thanks

test/ec_driver.cpp
```diff
@@ -582,31 +582,31 @@ int main(int argc, char** argv)
     data_zpad = count_digits(n_data - 1);
 
     if (eflag == EC_TYPE_RS_FNT) {
-        if (word_size <= 4) {
-            run_fec_rs_fnt<uint32_t>(
+        if (word_size == 1) {
```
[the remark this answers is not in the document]
lamphamsy: okay, I will update it

src/simd_fnt.h
lamphamsy: okay, it's better

src/simd_radix2_fft.h
lamphamsy: Thanks. I will update it

test/quadiron_c_utest.cpp
lamphamsy: yes, it's better. A copy of encoded fragments will be used to avoid such cases.

CMakeLists.txt
lamphamsy: Yes, it works now :)

lamphamsy:
@slaperche-scality : it's updated. Thanks for your next reviews :)

test/ec_driver.cpp
```diff
@@ -51,6 +51,13 @@ static void xusage()
     std::exit(EXIT_FAILURE);
 }
 
+[[noreturn]]
+static void xfnt()
+{
+    std::cerr << std::string("FEC using FNT supports word_size = 1 or 2\n");
```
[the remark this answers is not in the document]
lamphamsy: oh, you're right. Thanks

test/ec_driver.cpp
```diff
@@ -51,6 +51,13 @@ static void xusage()
     std::exit(EXIT_FAILURE);
 }
 
+[[noreturn]]
+static void xfnt()
```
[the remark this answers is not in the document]
lamphamsy: okay, I renamed it :)

test/ec_driver.cpp
```diff
@@ -52,9 +52,9 @@ static void xusage()
 }
 
 [[noreturn]]
-static void xfnt()
+static void xfnt_support_cases()
```
slaperche-scality: `xfnt_bad_word_size`?

lamphamsy: it's clearer. Thanks

PR #290, opened 2020-01-20
Add PDF documentation

vrancurel:


NicolasT:
Why not import the source? :man_shrugging:

vrancurel:
> Why not import the source? 🤷‍♂

done

